PR #6, opened 2019-04-07
Investigate FSM based parsing

Licenser:
The original code used a state machine to parse the JSON, this PR re-integrates it to see if we can get a speedup on duplicating this approach over doing it via recursion.


Licenser:
closing this as the 0.2 PR effectively implements it by parsing directly to tape.

PR #23, opened 2019-06-15
Experiment: re-use input buffer to store as much data as possible

Licenser:


Licenser:
The allocation test shows this would only be a minimal improvement and likely not worth it.

PR #26, opened 2019-06-23
Update README.md

DasEtwas:
Seems like I was squinting before, as I found another bunch of typos.
I also changed everything to say "simdjson-rs" as it's the repo's name.
The wording was kind of hacky so I polished it a bit.


DasEtwas:
The check seems nonsensical

Licenser:
Ja, I will ignore the checks, I was trying things out ;) thansk again!

PR #32, opened 2019-07-30
ARM NEON support

Licenser:

PR #35, opened 2019-08-08
RFC: Neon support (pretty much working)

sunnygleason:
Hello hello!

I have been pulling some of your Neon intrinsics and porting the
simdjson neon code. Maybe it's useful! I'll keep improving it...
Comments welcome anytime!

All the best,

-Sunny


Cargo.toml
```diff
@@ -13,6 +13,7 @@ halfbrown = "0.1"
 page_size = "0.4"
 itoa = "0.4"
 ryu = "1"
+core_arch = "0.1.5"
```
sunnygleason: not really necessary, just using it temporarily (also locally patched until it's released again - there is a known issue where mod.rs for core_arch needs to be patched to include feature, I think `stmt_expr_attributes`)

src/neon/intrinsics.rs
```diff
@@ -0,0 +1,287 @@
+//use std::arch::
+
+use crate::neon::simd_llvm;
+use crate::neon::simd;
+
```
sunnygleason: these have been gratuitously copy&pasted in...

src/neon/simd.rs
```diff
@@ -0,0 +1,469 @@
+#![allow(non_camel_case_types)]
```
sunnygleason: TODO: this file will be obsoleted by the core intrinsics work

src/neon/intrinsics.rs
```diff
@@ -0,0 +1,287 @@
+//use std::arch::
```
sunnygleason: TODO: this file will be obsoleted by the core intrinsics work

src/neon/simd_llvm.rs
```diff
@@ -0,0 +1,54 @@
+extern "platform-intrinsic" {
```
sunnygleason: TODO: this file will be obsoleted by the core intrinsics work

src/neon/utf8check.rs
```diff
@@ -0,0 +1,247 @@
+use crate::*;
```
sunnygleason: TODO: this file isn't neon yet

src/portability.rs
```diff
@@ -4,6 +4,7 @@ use std::arch::x86::*;
 use std::arch::x86_64::*;
 
 #[cfg_attr(not(feature = "no-inline"), inline)]
```
sunnygleason: TODO: this file isn't neon yet

src/value/generator.rs
```diff
@@ -4,7 +4,7 @@
 //
```
sunnygleason: TODO: this file isn't neon yet

src/numberparse.rs
```diff
@@ -133,26 +129,14 @@ pub enum Number {
 }
 
 #[cfg_attr(not(feature = "no-inline"), inline)]
+#[cfg(target_arch = "aarch64")]
```
Licenser: We’d probably want to move number parse in a x86 module and create a corresponding arm one instead of changing it to non simd code to cover both cases

Licenser:
That’s a great start!

Licenser:
We should have all the intrinsics (somewhat working) now! :D some are still brokenish however

src/neon/deser.rs
```diff
@@ -0,0 +1,204 @@
...
+            let bs_bits: u32 = 0; // FIXME
+//                unsafe {
+//                static_cast_u32!(_mm_movemask_epi8(vceqq_s8(
+//                    v,
+//                    vdupq_n_s8(b'\\' as i8)
+//                )))
+//            };
+            let quote_mask = 0; // FIXME: unsafe { vceqq_s8(v, vdupq_n_s8(b'"' as i8)) };
+            let quote_bits = 0; // FIXME: unsafe { static_cast_u32!(_mm_movemask_epi8(quote_mask)) };
+            if (bs_bits.wrapping_sub(1) & quote_bits) != 0 {
```
sunnygleason: Hm... @Licenser would you happen to know the ARM idiom for this movemask sequence?

src/neon/deser.rs
```diff
@@ -0,0 +1,204 @@
...
+
+        let src: &[u8] = unsafe { &self.input.get_unchecked(idx..) };
+        let mut src_i: usize = 0;
+        let mut len = src_i;
+        loop {
+            let v: int8x16_t = if src.len() >= src_i + 16 {
+                // This is safe since we ensure src is at least 16 wide
+                #[allow(clippy::cast_ptr_alignment)]
+                    unsafe {
+                    *(src.as_ptr().add(src_i) as *const int8x16_t)
```
sunnygleason: these were supposed to be load instructions

src/neon/intrinsics.rs
```diff
@@ -0,0 +1,395 @@
...
+    fn vtstq_u8_(a: uint8x16_t, b: uint8x16_t) -> uint8x16_t;
+    #[link_name = "llvm.aarch64.neon.vtstq.v16s8"]
+    fn vtstq_s8_(a: int8x16_t, b: int8x16_t) -> int8x16_t;
+    #[link_name = "llvm.ctpop.u64"]
+    fn ctpop_u64_(a: u64) -> u32;
+    #[link_name = "llvm.ctlz.u64"]
+    fn ctlz_u64_(a: u64) -> u32;
+    #[link_name = "llvm.cttz.u64"]
+    fn cttz_u64_(a: u64) -> u32;
+}
```
sunnygleason: haven't tested many of the above ... not completely sure they're right (or even close)

src/value/generator.rs
```diff
@@ -102,91 +102,91 @@ pub trait BaseGenerator {
...
             // The case where we have a 16+ byte block
             // we repeate the same logic as above but with
             // only 16 bytes
-            let zero = _mm_set1_epi8(0);
-            let lower_quote_range = _mm_set1_epi8(0x1F as i8);
-            let quote = _mm_set1_epi8(b'"' as i8);
-            let backslash = _mm_set1_epi8(b'\\' as i8);
-            while len - idx > 16 {
-                // Load 16 bytes of data;
-                #[allow(clippy::cast_ptr_alignment)]
```
sunnygleason: TODO: figure out the neon equivalent idiom

sunnygleason:
@Licenser I've uploaded some more work -- looking forward to your feedback! The intrinsics are mostly self-contained in `intrinsics.rs`; if you see any corrections/substitutions I can make, that would be awesome. I've mostly been just trying to get things to compile. I feel like we're getting pretty close once the intrinsics are valid!

src/neon/deser.rs
Licenser: Looking at the upstream code https://github.com/lemire/simdjson/blob/d9a0e2b8f441c20ad46276fdb8ce24f2aebdc07b/include/simdjson/stringparsing_arm64.h#L40 those two lines seem to be about setting the quote bits and mask

src/neon/deser.rs
Licenser: `vld1q_s8` should do the trick here?

src/neon/intrinsics.rs
Licenser: Do we need those? I went over the simdjson code and tried to find all the intrinsics used but I might have missed some / or things changed. If any are missing you can add them to the list in #32 and I'll add them :)

Licenser:
Did a quick pass over your comments :) it's a bit late here so I'll take a closer look tomorrow.

sunnygleason:
@Licenser I think we're getting closer. My main issue now is how to get around these errors:

```
`error: `extern` block uses type `neon::intrinsics::int8x16_t` which is not FFI-safe: this struct has unspecified layout
```

Overall, seems like we're getting closer...

src/neon/intrinsics.rs
sunnygleason: @Licenser I think so, unless there's equivalent simd_llvm idiom for hamming() and trailingzeroes()?

sunnygleason:
@Licenser ok, have it mostly building. Now I'm hitting issues in the compiler/link step I believe:

```
LLVM ERROR: Do not know how to scalarize the result of this operator!
```

src/neon/intrinsics.rs
```diff
@@ -301,7 +407,7 @@ pub fn zerou8x16() -> uint8x16_t {
 
 #[inline]
 pub unsafe fn vpaddq_u8(a: uint8x16_t, b: uint8x16_t) -> uint8x16_t {
-    mem::transmute(vpaddq_u8_(mem::transmute(a), mem::transmute(b)))
+    mem::transmute(vaddq_u8_(mem::transmute(a), mem::transmute(b)))
```
Licenser: addq and paddq are different intrinsics! This will lead to wrong results

Licenser:
So on a aarch64 cpu this compiles!

sunnygleason:
@Licenser I think I'm at a point where I could use some help. I've cross-checked the implementation with the C++ version -- things seem mostly ok, but still only 21 tests pass. It would be great to have another set of eyes check this out! Thank you so much for your help.

src/neon/intrinsics.rs
sunnygleason: ty!

Licenser:
Sure! There is a real chance that there is a bug in the intrinsics, who knows if I messed something up there :P it's not merged yet and for good reasons. 

So I did a bit of the good old println debugging (or `dbg!` :P). It looks like things go bananas in `flatten_bits`.

With the input `[1]` (`count2` test) the input and output is the following:

Before the call:
```
[src/neon/stage1.rs:521] &structurals = 7
[src/neon/stage1.rs:522] &idx = 64
[src/neon/stage1.rs:523] &structural_indexes = [
    0,
]
```
structurals is 7 (correct) 4 + 2 + 1 (first three bits, the first 3 characters are structurals, that's good!)

index: 64 since we had input length = 3 and we add 64 (we process that at a time) to the input (this might be where it geos wrong? but it's the same on avx and sse, so it should be right) 

After the call:
```
[src/neon/stage1.rs:525] &structurals = 7
[src/neon/stage1.rs:526] &idx = 64
[src/neon/stage1.rs:527] &structural_indexes = [
    0,
    64,
    2,
    1,
]
```

this is after the call, indexes are 0, 64, 2 and 1. I suspect that is where we go oops, it should be 0, 1, 2, 3. I'll take a peak at `flatten_bits` and the intrinsics in it.

sunnygleason:
@Licenser that makes sense, I'm seeing the same thing - bad character at index 64. I have augmented my code with println debugging so I will be able to get back to it later this evening -- in the meantime, thank you so much for your help!  :)

Licenser:
Not everything is working but the following diff makes quite a difference:
```diff
diff --git a/src/neon/stage1.rs b/src/neon/stage1.rs
index a1e15a1..2af15e6 100644
--- a/src/neon/stage1.rs
+++ b/src/neon/stage1.rs
@@ -332,7 +332,7 @@ unsafe fn flatten_bits(base: &mut Vec<u32>, idx: u32, mut bits: u64) {
         let v3: i32 = static_cast_i32!(trailingzeroes(bits));
         bits &= bits.wrapping_sub(1);

-        let v: int32x4_t = int32x4_t::new(v3, v2, v1, v0);
+        let v: int32x4_t = int32x4_t::new(v0, v1, v2, v3);
         let v: int32x4_t = vaddq_s32(idx_64_v, v);

         std::ptr::write(base.as_mut_ptr().add(l) as *mut int32x4_t, v);
```

Licenser:
here is another one:
```diff
diff --git a/src/numberparse.rs b/src/numberparse.rs
index 425d16a..b44d320 100644
--- a/src/numberparse.rs
+++ b/src/numberparse.rs
@@ -133,10 +133,10 @@ pub enum Number {
 fn parse_eight_digits_unrolled(chars: &[u8]) -> u32 {
     let val: u64 = unsafe { *(chars.as_ptr() as *const u64) };
     //    memcpy(&val, chars, sizeof(u64));
-    let val = (val & 0x0F0F0F0F0F0F0F0F) * 2561 >> 8;
-    let val = (val & 0x00FF00FF00FF00FF) * 6553601 >> 16;
+    let val = (val & 0x0F0F0F0F0F0F0F0F).wrapping_mul(2561) >> 8;
+    let val = (val & 0x00FF00FF00FF00FF).wrapping_mul(6553601) >> 16;

-    return ((val & 0x0000FFFF0000FFFF) * 42949672960001 >> 32) as u32;
+    return ((val & 0x0000FFFF0000FFFF).wrapping_mul(42949672960001) >> 32) as u32;
 }

 impl<'de> Deserializer<'de> {
```

With those two we're up to 66 passing tests.

That said we shouldn't generally fall back to the slow parse_eight_digits_unrolled, it should be architecture specific I think.

Licenser:
The remaining failing tests are:
```
    tests::crazy_string
    tests::map2
    tests::odd_array2
    tests::odd_array3
    tests::odd_array4
    tests::prop_json
    tests::prop_json_encode_decode
    tests::unicode
    tests::utf8
```

I gut feeling says it's unicode/utf8 related as all ood_* cases also include silly UTF8 strings.

sunnygleason:
That's awesome. The utf8 stuff has a lot of weirdness with 'u8' vs 'i8' vs 's8' and reinterpretation, so I'm not totally surprised. It could be reinterpretation issue, or me slightly modifying intrinsic(s). Pretty awesome to be at 66 passing tests, nice work!

sunnygleason:
(Also, I can confirm the 66 passing tests in my ARM environment)

Licenser:
All credit to you on the work I just swapped a `*` with a `wrapping_mul` (rust trivia: that would have worked in a release build AFAIK overflow checks are only enabled in dev / test builds) and changed the order of 4 variables :P I got the easy part!

sunnygleason:
OK, this is getting slightly closer, I think I messed up padding earlier. It might have to do with endian-ness? I think this latest patch provides more info. Any ideas?

sunnygleason:
Oh! I think I'm close - it seems to be an issue with how null bytes are handled?

image

sunnygleason:
Any thoughts?

Licenser:
The most basic test case I see failing is this:
```
---- tests::utf8 stdout ----
thread 'tests::utf8' panicked at 'assertion failed: `(left == right)`
  left: `String("\"")`,
 right: `"\u{e}"`', src/lib.rs:637:9
```

that could be either a copy or off by one error. I'll dig a little

Licenser:
So for the input `"\u000e"` 

```rust
 let ParseStringHelper { bs_bits, quote_bits } = unsafe { find_bs_bits_and_quote_bits(&srcx, None) };
```

returns

```
[src/neon/deser.rs:85] &(bs_bits, quote_bits) = (
    1,
    64,
)
```
i think the quote bits should be 0 I think (will need to double check) but the 64 on the backslash seems definetly off.

sunnygleason:
OK, so a little hacking and slashing makes me think maybe the store intrinsics aren't working as expected? (see latest commit). There is still one failing test, `odd_array2`, and the value-based tests for some reason...

src/neon/deser.rs
```diff
@@ -17,9 +18,11 @@ unsafe fn find_bs_bits_and_quote_bits(src: &[u8], dstx: Option<&mut [u8]>) -> Pa
     let v1 : uint8x16_t = vld1q_u8(src.as_ptr().add(16));
 
     match dstx {
```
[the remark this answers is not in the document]
Licenser: sorry I was wrong about that.

Licenser:
The following reduces it to 4 errors. It moves the copying in the string parse instead of the find bits and also gets rid of the nasty Option there.
```diff
diff --git a/src/lib.rs b/src/lib.rs
index f9ddb97..9f095f4 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -181,7 +181,10 @@ impl<'de> Deserializer<'de> {

         let counts = Deserializer::validate(input, &structural_indexes)?;

-        let strings = Vec::with_capacity(len + SIMDJSON_PADDING);
+        let mut strings = Vec::with_capacity(len + SIMDJSON_PADDING);
+        unsafe{
+            strings.set_len(len + SIMDJSON_PADDING);
+        }

         Ok(Deserializer {
             counts,
diff --git a/src/neon/deser.rs b/src/neon/deser.rs
index 65f744a..f422b3a 100644
--- a/src/neon/deser.rs
+++ b/src/neon/deser.rs
@@ -10,20 +10,12 @@ pub use crate::stringparse::*;

 pub use crate::neon::intrinsics::*;

-unsafe fn find_bs_bits_and_quote_bits(src: &[u8], dstx: Option<&mut [u8]>) -> ParseStringHelper {
+unsafe fn find_bs_bits_and_quote_bits(src: &[u8]) -> ParseStringHelper {
     // this can read up to 31 bytes beyond the buffer size, but we require
     // SIMDJSON_PADDING of padding
     let v0 : uint8x16_t = vld1q_u8(src.as_ptr());
     let v1 : uint8x16_t = vld1q_u8(src.as_ptr().add(16));

-    match dstx {
-        Some(dst) => {
-            vst1q_u8(dst.as_mut_ptr(), v0);
-            vst1q_u8(dst.as_mut_ptr().add(16), v1);
-        },
-        _ => ()
-    }
-
     let bs_mask : uint8x16_t = vmovq_n_u8('\\' as u8);
     let qt_mask : uint8x16_t = vmovq_n_u8('"' as u8);

@@ -81,7 +73,7 @@ impl<'de> Deserializer<'de> {
                 }
             };

-            let ParseStringHelper { bs_bits, quote_bits } = unsafe { find_bs_bits_and_quote_bits(&srcx, None) };
+            let ParseStringHelper { bs_bits, quote_bits } = unsafe { find_bs_bits_and_quote_bits(&srcx) };

             if (bs_bits.wrapping_sub(1) & quote_bits) != 0 {
                 // we encountered quotes first. Move dst to point to quotes and exit
@@ -120,7 +112,7 @@ impl<'de> Deserializer<'de> {
         }

         let mut dst_i: usize = 0;
-        let dst: &mut [u8] = &mut self.strings;
+        let dst = self.strings.as_mut_slice();

         loop {
             let srcx = if src.len() >= src_i + 32 {
@@ -133,10 +125,11 @@ impl<'de> Deserializer<'de> {
                     &padding
                 }
             };
+            dst[dst_i..dst_i+32].copy_from_slice(&srcx[..32]);

             // store to dest unconditionally - we can overwrite the bits we don't like
             // later
-            let ParseStringHelper { bs_bits, quote_bits } = unsafe { find_bs_bits_and_quote_bits(&srcx, Some(dst)) };
+            let ParseStringHelper { bs_bits, quote_bits } = unsafe { find_bs_bits_and_quote_bits(&srcx) };

             if (bs_bits.wrapping_sub(1) & quote_bits) != 0 {
                 // we encountered quotes first. Move dst to point to quotes and exit
@@ -213,4 +206,4 @@ impl<'de> Deserializer<'de> {
             }
         }
     }
-}
\ No newline at end of file
+}
```

sunnygleason:
@Licenser I think the issue is with the ARM pmull64 intrinsic. On aarch64, I'm seeing
bits in front of the mask that look possibly like sign extension. On x86, the mask
has no leading 1 bits. Do you have any ideas to remedy this?

(arm is unhappy, see QTM line)

arm_pmull64_issue

(sse42 is ok, see QTM line)

sse42_multiply_ok

Licenser:
So I'm a bit stumped at the last few remaining butgs, I was looking at the unicode test and it seems the escape in it isn't recognized.

Licenser:
ah damn GH not updating :/ sorry I just now saw you came to the same conclusion.  

the pmull intrinsic is a tricky one, I've not writte a test for it since from the manual (https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vmull_p64) I don't fully understand what it's supposed to do.

That said it does link to a llvm intrinsic so has no custom code but it could be the wrong intrinsic that it links against?

your screenshots are nicer then my printf debugging :joy: so I'll just comment on them.

To test I've ported the 'portable / non simd' version:
```rust
fn compute_quote_mask(quote_bits: u64) -> u64 {
  let quote_mask = quote_bits ^ (quote_bits << 1);
  let quote_mask = quote_mask ^ (quote_mask << 2);
  let quote_mask = quote_mask ^ (quote_mask << 4);
  let quote_mask = quote_mask ^ (quote_mask << 8);
  let quote_mask = quote_mask ^ (quote_mask << 16);
  quote_mask ^ (quote_mask << 32)
}
```

but it seems to have the same result on my tests, the backslash is not counted correctly. I have the feeling we're missing something trivial.

Licenser:
looking at it step by step things go off with odd_carry_ends I think:

```
backslashes:                                                       1000001000
backslashes:                                                       1000001000
even_carries:                                                       1000001000
bs_bits:                                                                1000001000
even_carries:                                                       1000001000
even_carry_ends:                                                                   0
```

sunnygleason:
@Licenser yes! I think we found that at the same time. Running tests now, maybe this is it!

Licenser:
Nice find! that makes all the basic tests pass :D so for me the jsonchecker tests now fail :(

Licenser:
update: `fail34` is the offending test at the moment.

Licenser:
I think that's the only left failing :D

Licenser:
So a little digging, it fails because the it doesn't fail on invalid utf8.

Licenser:
So I your code highlighted a bug in the intrinsics. They returned the wrong type which made the code need a lot more `vreinterpret` calls. If you want to pull them in that might clean up some of the issues.

sunnygleason:
@Licenser you're absolutely correct, thank you! Tests are passing now, check it out!

I guess next step is cleanup/beautification? ... Looking forward to your thoughts in that area!

Licenser:
Damn nice work sunny! :D and good call on copying the intrinsics over to not be blocked in the upstream MR!

I'll go over it and put in some comments / thoughts :)

sunnygleason:
Awesome, and compliments right back at you -- thanks for having the
bravery to dive into implementing the intrinsics in the first place! :)

Licenser:
Quick thing: can you add the following diff to the drone.yaml:

```diff
diff --git a/.drone.yml b/.drone.yml
index 9b8e8bb..8ceef40 100644
--- a/.drone.yml
+++ b/.drone.yml
@@ -58,5 +58,5 @@ steps:
 - name: test
   image: rust:1
   commands:
-  - cargo build --verbose --all
-  - cargo test --verbose --all
+  - cargo +nightly build --verbose --all
+  - cargo +nightly test --verbose --all
```

we'll need nightly for the tests so we might see them passing

src/neon/deser.rs
```diff
@@ -0,0 +1,210 @@
...
+                    &padding
+                }
+            };
+
+            let ParseStringHelper { bs_bits, quote_bits } = unsafe { find_bs_bits_and_quote_bits(&srcx) };
+
+            if (bs_bits.wrapping_sub(1) & quote_bits) != 0 {
+                // we encountered quotes first. Move dst to point to quotes and exit
+                // find out where the quote is...
+                let quote_dist: u32 = trailingzeroes(u64::from(quote_bits)) as u32;
```
Licenser: While looking at this I found that we can just use `.trailing_zeros()` here. It doesn't make sense to change it everywhere in this PR as it would conflate concerns, but we might as well start with the arm code to clean up some of those things :)

src/neon/deser.rs
```diff
@@ -0,0 +1,210 @@
...
+                len += 32;
+            }
+        }
+
+        let mut dst_i: usize = 0;
+        let dst: &mut [u8] = self.strings.as_mut_slice();
+
+        loop {
+            let srcx = if src.len() >= src_i + 32 {
+                &src[src_i..]
```
Licenser: we probably want `get_unchecked` here to prevent range checks on every iteration.

src/neon/deser.rs
```diff
@@ -0,0 +1,210 @@
...
+            } else {
+                unsafe {
+                    padding
+                        .get_unchecked_mut(..src.len() - src_i)
+                        .clone_from_slice(src.get_unchecked(src_i..));
+                    &padding
+                }
+            };
+
+            dst[dst_i..dst_i + 32].copy_from_slice(&srcx[..32]);
```
Licenser: Again, my fault here, but this should be

```rust
unsafe {
  dst.get_unchecked_mut(dst_i..dst_i + 32).copy_from_slice(srcx.get_unchecked(..32));
}
```

to avoid checking the range on every iteration.

src/neon/deser.rs
```diff
@@ -0,0 +1,210 @@
...
+
+pub use crate::neon::intrinsics::*;
+
+unsafe fn find_bs_bits_and_quote_bits(src: &[u8]) -> ParseStringHelper {
+    // this can read up to 31 bytes beyond the buffer size, but we require
+    // SIMDJSON_PADDING of padding
+    let v0 : uint8x16_t = vld1q_u8(src.as_ptr());
+    let v1 : uint8x16_t = vld1q_u8(src.as_ptr().add(16));
+
+    let bs_mask : uint8x16_t = vmovq_n_u8('\\' as u8);
```
Licenser: this can be written as `b'\\'` (same below)

sunnygleason:
@Licenser is it an easy tweak to add this to drone CI?

image

src/neon/deser.rs
Licenser: In arm this would also remove the need for conversion to u64 I believe.

src/neon/mod.rs
```diff
@@ -0,0 +1,6 @@
+pub mod deser;
+pub mod stage1;
+pub mod utf8check;
+mod simd;
```
Licenser: adding a note here so we don't forget to remove that prior to merge / once the intrinsics are in nightly

src/neon/stage1.rs
```diff
@@ -0,0 +1,531 @@
...
+            v1: vld1q_u8(ptr.as_ptr().add(16) as *const u8),
+            v2: vld1q_u8(ptr.as_ptr().add(32) as *const u8),
+            v3: vld1q_u8(ptr.as_ptr().add(48) as *const u8),
+        }
+    }
+}
+
+#[cfg_attr(not(feature = "no-inline"), inline(always))]
+unsafe fn neon_movemask(input: uint8x16_t) -> u16 {
+    let bit_mask: uint8x16_t = uint8x16_t::new(0x01, 0x02, 0x4, 0x8, 0x10, 0x20, 0x40, 0x80,
```
Licenser: We have this mask in a few places, perhaps we can `const` it out and use a load instruction to load it in a register? (I see it twice in this file and once in `deser.rs` I think)

src/neon/stage1.rs
```diff
@@ -0,0 +1,531 @@
...
+                                          &mut (state.previous), &mut (state.has_error));
+        state.previous = check_utf8_bytes(vreinterpretq_s8_u8(input.v3),
+                                          &mut (state.previous), &mut (state.has_error));
+    }
+}
+
+// a straightforward comparison of a mask against input
+#[cfg_attr(not(feature = "no-inline"), inline(always))]
+unsafe fn cmp_mask_against_input(input: &SimdInput, m: u8) -> u64 {
+    macro_rules! call {
```
Licenser: What's the story of the `constify_imm8` macro in here, are we missing an intrinsic?

src/neon/stage1.rs
```diff
@@ -0,0 +1,531 @@
...
+
+// find all values less than or equal than the content of maxval (using unsigned arithmetic)
+#[cfg_attr(not(feature = "no-inline"), inline(always))]
+unsafe fn unsigned_lteq_against_input(input: &SimdInput, maxval: u8) -> u64 {
+    macro_rules! call {
+        ($imm8:expr) => {
+            vmovq_n_u8($imm8)
+        };
+    };
+    let mask: uint8x16_t = constify_imm8!(maxval, call);
```
Licenser: What's the storry of the `constify_imm8` here, are we missing an intrinsic? I have the feeling a macro like that should never be called outside of stdlib/intrinsic code.

would:
```
let mask: uint8x16_t = vmovq_n_u8(maxval);
```
 do the trick?

src/neon/stage1.rs
```diff
@@ -0,0 +1,531 @@
...
+    let even_start_mask: u64 = EVEN_BITS ^ *prev_iter_ends_odd_backslash;
+    let even_starts: u64 = start_edges & even_start_mask;
+    let odd_starts: u64 = start_edges & !even_start_mask;
+    let even_carries: u64 = bs_bits.wrapping_add(even_starts);
+
+    let mut odd_carries: u64 = 0;
+    // must record the carry-out of our odd-carries out of bit 63; this
+    // indicates whether the sense of any edge going to the next iteration
+    // should be flipped
+    let iter_ends_odd_backslash: bool = add_overflow(bs_bits, odd_starts, &mut odd_carries);
```
Licenser: Could we just pull in the rust own function here, it might save us another architecture specific function the same way as `trailing_zeros()` does.

src/neon/stage1.rs
```diff
@@ -0,0 +1,531 @@
...
+}
+
+// flatten out values in 'bits' assuming that they are are to have values of idx
+// plus their position in the bitvector, and store these indexes at
+// base_ptr[base] incrementing base as we go
+// will potentially store extra values beyond end of valid bits, so base_ptr
+// needs to be large enough to handle this
+//TODO: usize was u32 here does this matter?
+#[cfg_attr(not(feature = "no-inline"), inline(always))]
+unsafe fn flatten_bits(base: &mut Vec<u32>, idx: u32, mut bits: u64) {
```
Licenser: The upstream code has gone to unroll this loop, I'll open a ticket (#37) to adopt that in a general ways  but since this code is still getting worked on we could start with it here?

Licenser: On looking at this, we have a slightly different appraoch, it might be worth comparing the two. and postpone this change for later after doing some benchmarks. Our implementation might be faster as it stands but could be even faster faster w/ unrolling. Feel free to discard if you feel like it doesn't belong here.

src/neon/utf8check.rs
```diff
@@ -0,0 +1,191 @@
...
+    // unsigned, saturates to 0 below max
+    *has_error =
+        vorrq_s8(
+            *has_error,
+            vqsubq_s8(current_bytes, vdupq_n_s8(-12 /* 0xF4 */)));
+}
+
+#[cfg_attr(not(feature = "no-inline"), inline)]
+unsafe fn continuation_lengths(high_nibbles: int8x16_t) -> int8x16_t {
+    let nibbles: int8x16_t = int8x16_t::new(
```
Licenser: I think upstream `consts` this out and used a load instruction to fetch it, perhaps we could/should do the same?

src/neon/utf8check.rs
```diff
@@ -0,0 +1,191 @@
...
+// else      false && false
+#[cfg_attr(not(feature = "no-inline"), inline)]
+unsafe fn check_overlong(
+    current_bytes: int8x16_t,
+    off1_current_bytes: int8x16_t,
+    hibits: int8x16_t,
+    previous_hibits: int8x16_t,
+    has_error: &mut int8x16_t,
+) {
+    let _initial_mins: int8x16_t = int8x16_t::new(
```
Licenser: since it's used the variable shouldn't start with an `_`

src/numberparse.rs
```diff
@@ -133,26 +129,14 @@ pub enum Number {
...
-            _mm_loadu_si128(chars.get_unchecked(0..16).as_ptr() as *const __m128i),
-            ascii0,
-        );
-        let t1: __m128i = _mm_maddubs_epi16(input, mul_1_10);
-        let t2: __m128i = _mm_madd_epi16(t1, mul_1_100);
-        let t3: __m128i = _mm_packus_epi32(t2, t2);
-        let t4: __m128i = _mm_madd_epi16(t3, mul_1_10000);
-        _mm_cvtsi128_si32(t4) as u32 // only captures the sum of the first 8 digits, drop the rest
-    }
+    let val: u64 = unsafe { *(chars.as_ptr() as *const u64) };
```
Licenser: we shouldn't fall back to the general "slow" implementation on all architectures, lets make this architecture dependant and keep the x86 version SIMD'd and the ARM version plain.

src/stage2.rs
```diff
@@ -1,9 +1,11 @@
 #![allow(dead_code)]
+use crate::charutils::*;
 #[cfg(target_feature = "avx2")]
 use crate::avx2::stage1::SIMDJSON_PADDING;
-use crate::charutils::*;
-#[cfg(not(target_feature = "avx2"))]
+#[cfg(target_feature = "sse4.2")]
```
Licenser: See comments about guarding in `lib.rs`

src/portability.rs
```diff
@@ -17,12 +18,13 @@ pub fn hamming(input_num: u64) -> u32 {
...
-#[cfg(target_arch = "x86")]
+#[cfg(target_arch = "x86_64")]
 pub fn hamming(input_num: u64) -> u32 {
     unsafe { __popcnt(input_num as u32) + __popcnt((input_num >> 32) as u32) as u32 }
 }
 
 #[cfg_attr(not(feature = "no-inline"), inline)]
+#[cfg(target_arch = "x86_64")]
```
Licenser: This should guard for both x86 and x86_64 architectures, right now it'd not compile on a 32 bit system.

src/stringparse.rs
```diff
@@ -73,3 +73,9 @@ pub fn handle_unicode_codepoint(
     let offset: usize = codepoint_to_utf8(code_point, dst_ptr);
     Ok((offset, src_offset))
 }
+
+// Holds backslashes and quotes locations.
+pub struct ParseStringHelper {
```
Licenser: I think this is arm specific isn't it? If so it should live in the arm sub directory.

src/value/generator.rs
```diff
@@ -98,95 +98,96 @@ pub trait BaseGenerator {
...
-        let mut string = string.as_bytes();
-        let mut len = string.len();
-        let mut idx = 0;
+        let string = string.as_bytes();
+//        let mut string = string.as_bytes();
+//        let mut len = string.len();
+//        let mut idx = 0;
 
-        unsafe {
+//        unsafe {
```
Licenser: this should be commented in again. a simple solution would be to guard the unsafe with a 

```rust
#[cfg(target_feature = "avx2")]
unsafe {
...
}
```

That way we don't need compile time checks at all and the code only gets included when the right intrinsics are present.

Licenser:
darn, it seems we need another image this might need a thing or two of experimentation, lets try

```diff
libdidiff --git a/.drone.yml b/.drone.yml
index 9b8e8bb..74813b3 100644
--- a/.drone.yml
+++ b/.drone.yml
@@ -56,7 +56,7 @@ platform:

 steps:
 - name: test
-  image: rust:1
+  image: rustlang/rust:nightly
   commands:
   - cargo build --verbose --all
   - cargo test --verbose --all
```

If that does't work we might need a combination of both but I don't think we do. Sorry for that oversight.

sunnygleason:
Hm, not quite. Maybe we could add a `rustup install nightly` command to the previous config? Or any idea you have, which is probably better  :)

image

Licenser:
Good idea, the error seems to hint to the fact that the image doesn't exist for arm, so perhaps back to the library image (`rust:1`) and as you suggest do a rustup install nightly on that. 

Sadly the "lets get the CI play ball" part is just about trying random combinations of YAML until it works 😂

src/neon/deser.rs
sunnygleason: @Licenser maybe I'll sneak a PR in ahead of this for the `.trailing_zeros()` usage on the x86 side?

Licenser: sounds good! 👍

src/neon/utf8check.rs
sunnygleason: sounds good - I was running into some issues with the `vld*` instruction, will pick away at it...

src/stringparse.rs
sunnygleason: @Licenser one thing that's neat about the `find_quote_mask_and_bits()` function is that it encapsulates platform-specific logic in the main deserializer function, maybe we should consider adopting the pattern on the x86 side? (Which will slim down code...)

sunnygleason:
@Licenser drone CI works! (for arm64) 🎉

sunnygleason:
@Licenser what's the best way to handle that portability commit - would you be able to rebase your `arm` branch?

Licenser:
Was about todo that :D

src/stringparse.rs
sunnygleason: @Licenser the code between the architectures is converging -- I have a nifty refactoring that I'll send as another PR.

sunnygleason:
OK, this should be better!

* let me know what the strategy should be re: nightly & merging
* arch refactor preview here : https://github.com/sunnygleason/simdjson-rs/pull/1/files
* I'll stay tuned for any other comments/questions/concerns, thanks again!

src/neon/stage1.rs
sunnygleason: I'd love to bring it in from upstream soon as part of some arch consolidation work!

src/neon/stage1.rs
```diff
@@ -0,0 +1,603 @@
...
+    fn default() -> Self {
+        Utf8CheckingState {
+            has_error: vdupq_n_s8(0),
+            previous: ProcessedUtfBytes::default(),
+        }
+    }
+}
+
+#[inline]
+fn is_utf8_status_ok(has_error: int8x16_t) -> bool {
```
Licenser: I'm a bit confused about this. 'AND' the data with itself, which has no effect and then turn it into a 128 bit type and then into a 16 bit type and then compare it to zero.

What is it we actually trying to do? Check that everything i there is zero?

src/value/generator.rs
```diff
@@ -5,19 +5,30 @@
...
+#[cfg(any(target_feature = "sse4.2", target_feature = "avx2"))]
+use std::arch::x86_64::*;
+#[cfg(target_feature = "neon")]
+use crate::*;
+
 #[cfg(target_feature = "avx2")]
 const AVX2_PRESENT : bool = true;
 #[cfg(not(target_feature = "avx2"))]
 const AVX2_PRESENT : bool = false;
+#[cfg(target_feature = "sse4.2")]
```
Licenser: those are not needed, we don't need to do runtime checks. you can simply put the whole thing in a block and guard the block with a `#[cfg]` directive :) that'll prevent us from having to do it over and over again. 

The compiler might (should) optimize a if false away but having the `#[cfg]` guard around the code that isn't executed makes it clearer when reading the code that it's something that happens at compile not at runtime.

src/value/generator.rs
```diff
@@ -107,84 +118,11 @@ pub trait BaseGenerator {
...
-                        escape => stry!(self.write(&[b'\\', escape])),
-                    };
-                    string = &string[idx + quote_dist + 1..];
-                    idx = 0;
-                    len = string.len();
-                } else {
-                    idx += 16;
-                }
+            if SSE42_PRESENT || NEON_PRESENT {
+                self.process_16_bytes(&mut string, &mut len, &mut idx).unwrap();
```
Licenser: we can't have unwrap in library code. this would possibly lead to the program terminating because of bad input / a bug and library users can't handle the failure racefully.

this should be `stry!(...)`

src/value/generator.rs
```diff
@@ -197,6 +135,170 @@ pub trait BaseGenerator {
         self.write_char(b'"')
     }
 
+    #[cfg(target_feature = "neon")]
```
Licenser: this should live somewhere in the `neon/` directory. Or more generally since there are now 3 different implementation of write_string they all should be in their respective directories.

src/value/generator.rs
```diff
@@ -197,6 +135,170 @@ pub trait BaseGenerator {
...
+            } else {
+                *idx += 16;
+            }
+        }
+        stry!(self.get_writer().write_all(&string[0..*idx]));
+        *string = &string[*idx..];
+        Ok(())
+    }
+
+    #[cfg(all(not(target_feature = "sse4.2"), not(target_feature = "neon")))]
```
Licenser: With either guarding the the code in the function or better locating the right function into the right sub directly those stubs are no longer needed.

src/neon/stage1.rs
sunnygleason: yeah that was confusing to me too, it's widening over the whole bit width... return 1 if AND(a, b) is all zero bits

Licenser: So `vandq_s16` is bitewise and, not logical [1] so `vandq_s16(has_error, has_error) == has_error`, then we pull it out as a single 128b integer and extract only the last 16 (so in the case the last hightest), ignoring 1-3 all others.

I'm worried that this passes our test cases but is logically incorrect, at least all I can come up with screams 'this is wrong'.

Partially because A && A seems obsolute, and we're down-casting and throwing away a lot of bits.


[1] https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vandq_s16

Licenser: So what also worries me is that if I remove the and in there all tests pass, so either it really has no effect or the tests do not cover any case where it does.

And even with the  and back in there I could construct a test case that panics this. (fail34 with the bad utf8 moved up forward)

Licenser:
```diff
diff --git a/data/fail/fail34.json b/data/fail/fail34.json
index e82c700..0445278 100644
--- a/data/fail/fail34.json
+++ b/data/fail/fail34.json
@@ -1 +1 @@
-["this string contains bad UTF-8 <80>"]
+[" <80>"]
diff --git a/src/neon/stage1.rs b/src/neon/stage1.rs
index b23f059..65e2535 100644
--- a/src/neon/stage1.rs
+++ b/src/neon/stage1.rs
@@ -114,13 +114,10 @@ impl Default for Utf8CheckingState {
 fn is_utf8_status_ok(has_error: int8x16_t) -> bool {
     unsafe {
         let utf8_error_bits: u128 = mem::transmute(
-            vandq_s16(
-                mem::transmute(has_error),
-                mem::transmute(has_error)
-            )
+            has_error
         );

-        utf8_error_bits as u16 == 0
+        utf8_error_bits == 0
     }
 }
```

This passes correctly for.

I'll add some fail cases that cover 'bad' utf8 in different places of the input, this just slipped through since we only had it in one location and that happend  to fall into the 16 bit.

sunnygleason: @Licenser great catch!

src/value/generator.rs
sunnygleason: updated to use macros; the trait modeling is slightly beyond my experience level since we're going outside the current module, advice welcome!

sunnygleason:
@Licenser ok this is weird, arm64 passes for me in EC2 but not on drone. Are you able to repro the test failure?

src/neon/stage1.rs
```diff
@@ -113,14 +113,9 @@ impl Default for Utf8CheckingState {
...
-        let utf8_error_bits: u128 = mem::transmute(
-            vandq_s16(
-                mem::transmute(has_error),
-                mem::transmute(has_error)
-            )
-        );
+        let has_error_128 : i128 = mem::transmute(has_error);
 
-        utf8_error_bits as u16 == 0
+        has_error_128 as u32 == 0
```
Licenser: I don't think any type cast should happen here this line should be
```rust
has_error_128  == 0
```
as has_error is a `int8x16_t` so 16*8 bit or 128 bit in total. downcasting that to anything else will ignore all the upper bits.

sunnygleason: success!

sunnygleason:
@Licenser success!

src/value/generator.rs
Licenser: I'll add a diff at the end of this but I'll walk through the theory first:

So rust is quite powerful when it comes to generics and the dependant compilation of whole module trees.

We'll use that to our advantage to make the implementation zero overhead (and I don't mean nearly zero overhead but just have it slot in where it is requred at no cost at runtime not even a nop'ed out funciton call).

1st: We call the function the same in all implementations. This allows us to use a dependant import the same way we use it for the stage1 and other modules.

2nd: We pull the implementation out of the trait and make it generic over the writer. Generic here doesn't mean dynamic dispatch. Rust will compile the function for each generic variant that is used. In our case we make it generic over the writer (called `W` in the implementation).

3rd: NEON case (we ca reuse the `neon_movemask` from stage1 - I think this eventually should go into a file of shared functions but I wanted to keep the diff small.

4th: follow the same step for sse and avx.

result:  No macros, no runtime overhead. Yay!

The whole rust generics, especially when mixed with dependant compilation and multiple architectures, take a some time to wrap your head around :D it's one of those things that just require you to do run in a wall 15000 times until it clicks!

```diff
diff --git a/src/neon/generator.rs b/src/neon/generator.rs
index 7e0c8b6..7c9eaad 100644
--- a/src/neon/generator.rs
+++ b/src/neon/generator.rs
@@ -1,32 +1,10 @@
+use crate::value::generator::ESCAPED;
+use std::io;
+use crate::neon::intrinsics::*;
+use crate::neon::stage1::neon_movemask;

-#[macro_export]
-macro_rules! process_32_bytes {
-    () => {
         #[inline(always)]
-        unsafe fn process_32_bytes(&mut self, _string: &mut &[u8], _len: &mut usize, _idx: &mut usize) -> io::Result < () > {
-            Ok(())
-        }
-    };
-}
-
-#[macro_export]
-macro_rules! process_16_bytes {
-    () => {
-        #[inline(always)]
-        unsafe fn process_16_bytes(&mut self, string: &mut &[u8], len: &mut usize, idx: &mut usize) -> io::Result<()> {
-            #[cfg_attr(not(feature = "no-inline"), inline(always))]
-            unsafe fn __neon_movemask(input: uint8x16_t) -> u16 {
-                let bit_mask = uint8x16_t::new(
-                    0x01, 0x02, 0x4, 0x8, 0x10, 0x20, 0x40, 0x80,
-                    0x01, 0x02, 0x4, 0x8, 0x10, 0x20, 0x40, 0x80
-                );
-                let minput: uint8x16_t = vandq_u8(input, bit_mask);
-                let tmp: uint8x16_t = vpaddq_u8(minput, minput);
-                let tmp = vpaddq_u8(tmp, tmp);
-                let tmp = vpaddq_u8(tmp, tmp);
-
-                vgetq_lane_u16(vreinterpretq_u16_u8(tmp), 0)
-            }
+        pub unsafe fn write_str_simd<W>(writer: &mut W, string: &mut &[u8], len: &mut usize, idx: &mut usize) -> io::Result<()> where W: std::io::Write {

             // The case where we have a 16+ byte block
             // we repeate the same logic as above but with
@@ -48,15 +26,15 @@ macro_rules! process_16_bytes {
                 // now.
                 let is_unchanged = vxorrq_u8(data, in_quote_range);
                 let in_range = vceqq_u8(is_unchanged, zero);
-                let quote_bits = __neon_movemask(vorrq_u8(bs_or_quote, in_range));
+                let quote_bits = neon_movemask(vorrq_u8(bs_or_quote, in_range));
                 if quote_bits != 0 {
                     let quote_dist = quote_bits.trailing_zeros() as usize;
-                    stry!(self.get_writer().write_all(&string[0..*idx + quote_dist]));
+                    stry!(writer.write_all(&string[0..*idx + quote_dist]));
                     let ch = string[*idx + quote_dist];
                     match ESCAPED[ch as usize] {
-                        b'u' => stry!(write!(self.get_writer(), "\\u{:04x}", ch)),
+                        b'u' => stry!(write!(writer, "\\u{:04x}", ch)),

-                        escape => stry!(self.write(&[b'\\', escape])),
+                        escape => stry!(writer.write_all(&[b'\\', escape])),
                     };
                     *string = &string[*idx + quote_dist + 1..];
                     *idx = 0;
@@ -65,9 +43,7 @@ macro_rules! process_16_bytes {
                     *idx += 16;
                 }
             }
-            stry!(self.get_writer().write_all(&string[0..*idx]));
+            stry!(writer.write_all(&string[0..*idx]));
             *string = &string[*idx..];
             Ok(())
         }
-    };
-}
\ No newline at end of file
diff --git a/src/neon/stage1.rs b/src/neon/stage1.rs
index 0f00857..45322ce 100644
--- a/src/neon/stage1.rs
+++ b/src/neon/stage1.rs
@@ -18,7 +18,7 @@ macro_rules! bit_mask {
 }

 #[cfg_attr(not(feature = "no-inline"), inline(always))]
-unsafe fn neon_movemask(input: uint8x16_t) -> u16 {
+pub(crate) unsafe fn neon_movemask(input: uint8x16_t) -> u16 {
     let minput: uint8x16_t = vandq_u8(input, bit_mask!());
     let tmp: uint8x16_t = vpaddq_u8(minput, minput);
     let tmp = vpaddq_u8(tmp, tmp);
@@ -28,7 +28,7 @@ unsafe fn neon_movemask(input: uint8x16_t) -> u16 {
 }

 #[cfg_attr(not(feature = "no-inline"), inline(always))]
-unsafe fn neon_movemask_bulk(p0: uint8x16_t, p1: uint8x16_t, p2: uint8x16_t, p3: uint8x16_t) -> u64 {
+pub unsafe fn neon_movemask_bulk(p0: uint8x16_t, p1: uint8x16_t, p2: uint8x16_t, p3: uint8x16_t) -> u64 {
     let bit_mask = bit_mask!();

     let t0 = vandq_u8(p0, bit_mask);
diff --git a/src/value.rs b/src/value.rs
index 0624587..c5d5aa8 100644
--- a/src/value.rs
+++ b/src/value.rs
@@ -11,7 +11,7 @@
 /// we do not require prior knowledge sbout string comtent to to take advantage
 /// of it.
 pub mod borrowed;
-mod generator;
+pub(crate) mod generator;
 pub mod owned;

 pub use self::borrowed::{to_value as to_borrowed_value, Value as BorrowedValue};
diff --git a/src/value/generator.rs b/src/value/generator.rs
index b053344..007acc1 100644
--- a/src/value/generator.rs
+++ b/src/value/generator.rs
@@ -12,19 +12,19 @@ use std::ptr;

 #[cfg(target_arch = "x86")]
 use std::arch::x86::*;
-#[cfg(any(target_feature = "sse4.2", target_feature = "avx2"))]
+#[cfg(target_arch = "x86_64")]
 use std::arch::x86_64::*;

 use crate::*;

 #[cfg(target_feature = "avx2")]
-pub use crate::avx2::generator::*;
+use crate::avx2::generator::*;

 #[cfg(all(any(target_arch = "x86", target_arch = "x86_64"), not(target_feature = "avx2")))]
-pub use crate::sse42::generator::*;
+use crate::sse42::generator::*;

 #[cfg(target_feature = "neon")]
-pub use crate::neon::generator::*;
+use crate::neon::generator::*;

 const QU: u8 = b'"';
 const BS: u8 = b'\\';
@@ -37,7 +37,7 @@ const UU: u8 = b'u';
 const __: u8 = 0;

 // Look up table for characters that need escaping in a product string
-static ESCAPED: [u8; 256] = [
+pub(crate) static ESCAPED: [u8; 256] = [
     // 0   1   2   3   4   5   6   7   8   9   A   B   C   D   E   F
     UU, UU, UU, UU, UU, UU, UU, UU, BB, TT, NN, UU, FF, RR, UU, UU, // 0
     UU, UU, UU, UU, UU, UU, UU, UU, UU, UU, UU, UU, UU, UU, UU, UU, // 1
@@ -113,8 +113,7 @@ pub trait BaseGenerator {
             // quote characters that gives us a bitmask of 0x1f for that
             // region, only quote (`"`) and backslash (`\`) are not in
             // this range.
-            stry!(self.process_32_bytes(&mut string, &mut len, &mut idx));
-            stry!(self.process_32_bytes(&mut string, &mut len, &mut idx));
+            stry!(write_str_simd(self.get_writer(), &mut string, &mut len, &mut idx));
         }
         // Legacy code to handle the remainder of the code
         for (index, ch) in string.iter().enumerate() {
@@ -127,12 +126,6 @@ pub trait BaseGenerator {
         self.write_char(b'"')
     }

-    // 32-byte generation implementation
-    process_32_bytes!();
-
-    // 16-byte generation implementation
-    process_16_bytes!();
-
     #[inline(always)]
     fn write_float(&mut self, num: f64) -> io::Result<()> {
         let mut buffer = ryu::Buffer::new();
```

Licenser:
We are getting there :D

Licenser:
Now I just need to get the last 5 intrinsics working :/ I split the MR in the hope that we can get 90% merged while I try to wrangle the remaining ones.

Licenser:
Now I just need to get the last 5 intrinsics working :/ I split the MR in the hope that we can get 90% merged while I try to wrangle the remaining ones.

src/value/generator.rs
sunnygleason: that is amazing 🚀  ... integrating!

src/value/generator.rs
```diff
@@ -5,19 +5,26 @@
...
-#[cfg(target_arch = "x86_64")]
-use std::arch::x86_64::*;
 use std::io;
 use std::io::Write;
 use std::marker::PhantomData;
 use std::ptr;
 
+#[cfg(target_arch = "x86")]
+use std::arch::x86::*;
+#[cfg(any(target_feature = "sse4.2", target_feature = "avx2"))]
```
Licenser: this should be `#[cfg(target_arch = "x86_64")]` otherwise it will be double imported on a 32bit OS w/ avs2 or see4.2

src/value/generator.rs
```diff
@@ -106,85 +113,8 @@ pub trait BaseGenerator {
...
-                    string = &string[idx + quote_dist + 1..];
-                    idx = 0;
-                    len = string.len();
-                } else {
-                    idx += 16;
-                }
-            }
-            stry!(self.get_writer().write_all(&string[0..idx]));
-            string = &string[idx..];
+            stry!(self.process_32_bytes(&mut string, &mut len, &mut idx));
```
Licenser: marker: see discussion about using a generic function.

src/sse42/generator.rs
```diff
@@ -0,0 +1,72 @@
+
+#[macro_export]
```
Licenser: marker: see discussion about using a generic function.

src/neon/generator.rs
```diff
@@ -0,0 +1,73 @@
+
+#[macro_export]
```
Licenser: marker: see discussion about using a generic function.

src/avx2/generator.rs
```diff
@@ -0,0 +1,97 @@
+
+#[macro_export]
```
Licenser: marker: see discussion about using a generic function.

sunnygleason:
@Licenser ok, hopefully it works! :)

PR #39, opened 2019-08-30
Use simd-lite

Licenser:
Use simd-lite for ARM support.


src/neon/generator.rs
```diff
@@ -1,7 +1,8 @@
 use crate::value::generator::ESCAPED;
 use std::io;
-use crate::neon::intrinsics::*;
 use crate::neon::stage1::neon_movemask;
+use std::arch::aarch64::*;
```
sunnygleason: this leaves a sour taste...

src/neon/stage1.rs
```diff
@@ -1,18 +1,19 @@
...
 use std::mem;
 
 // NEON-SPECIFIC
 
 macro_rules! bit_mask {
     () => {
-        uint8x16_t::new(
-            0x01, 0x02, 0x4, 0x8, 0x10, 0x20, 0x40, 0x80,
-            0x01, 0x02, 0x4, 0x8, 0x10, 0x20, 0x40, 0x80
+       mem::transmute::<[u8;16],uint8x16_t>(
```
sunnygleason: this leaves a sour taste, I'm thinking a developer shouldn't have to use transmute for simple day-to-day tasks

sunnygleason:
@Licenser it's awesome that this works! The developer opinions aren't a deal-breaker, I think we can fix that in future revisions anyway.

src/neon/stage1.rs
Licenser: you're right we should use a load instruction here! :)

sunnygleason:
@Licenser was just looking at the test failures, I think you might want to use this impl of vtstq. I did an integration earlier against your intrinsics work, and this is what I needed (which is why I was complaining about the private fields):

```

#[inline]
fn test_u8(a: u8, b: u8) -> u8 {
    if a & b != 0 {
        0xFF
    } else {
        0x00
    }
}

#[inline]
pub unsafe fn vtstq_u8(a: uint8x16_t, b: uint8x16_t) -> uint8x16_t {
    uint8x16_t(
        test_u8(a.0, b.0),
        test_u8(a.1, b.1),
        test_u8(a.2, b.2),
        test_u8(a.3, b.3),
        test_u8(a.4, b.4),
        test_u8(a.5, b.5),
        test_u8(a.6, b.6),
        test_u8(a.7, b.7),
        test_u8(a.8, b.8),
        test_u8(a.9, b.9),
        test_u8(a.10, b.10),
        test_u8(a.11, b.11),
        test_u8(a.12, b.12),
        test_u8(a.13, b.13),
        test_u8(a.14, b.14),
        test_u8(a.15, b.15),
    )
}

#[inline]
fn test_s8(a: i8, b: i8) -> i8 {
    if a & b != 0 {
        -1
    } else {
        0x00
    }
}

#[inline]
pub unsafe fn vtstq_s8(a: int8x16_t, b: int8x16_t) -> int8x16_t {
    int8x16_t(
        test_s8(a.0, b.0),
        test_s8(a.1, b.1),
        test_s8(a.2, b.2),
        test_s8(a.3, b.3),
        test_s8(a.4, b.4),
        test_s8(a.5, b.5),
        test_s8(a.6, b.6),
        test_s8(a.7, b.7),
        test_s8(a.8, b.8),
        test_s8(a.9, b.9),
        test_s8(a.10, b.10),
        test_s8(a.11, b.11),
        test_s8(a.12, b.12),
        test_s8(a.13, b.13),
        test_s8(a.14, b.14),
        test_s8(a.15, b.15),
    )
}
```

src/neon/generator.rs
Licenser: I kind of feel we don't want to re-implement it all, that'll make it really hard to use simd-lite with std and that feels just wrong.

Licenser:
```rust
/// Compare bitwise test bits nonzero
#[inline]
#[target_feature(enable = "neon")]
#[cfg_attr(target_arch = "arm", target_feature(enable = "v7"))]
#[cfg_attr(test, assert_instr(cmtst))]
// even gcc compiles this to ldr: https://clang.godbolt.org/z/1bvH2x
// #[cfg_attr(test, assert_instr(ld1))]
pub unsafe fn vtstq_u8(a: uint8x16_t, b: uint8x16_t) -> uint8x16_t {
    vcgtq_u8(vandq_u8(a, b), vdupq_n_u8(0))
}
```

Ja I think that should be the same as `vcgtq_u8(vandq_u8(a, b), vdupq_n_u8(0))`

`vandq_u8(a, b)` should be `a & b` (vectorized) and `vcgtq_u8(..., vdupq_n_u8(0))` should be `!= 0` (or `>= `) vectorized (I dind't find a not-equal intrinsic but perhaps I just missed it?)

sadly not even clang compiles `vtstq_u8` to a own intrinsic: https://godbolt.org/z/lwHshr

sunnygleason:
I'm crying 😂 

image

Licenser:
I know it's ugly :( but it's correcter then mem::transmute, it's the load intrinsic that's made to well ... populate a register.

Licenser:
ideally we want an neon equivalent of `_mm256_setr_epi8`

Licenser:
hmm I got an idea! We could use a trait to add a ::new() method how about that?

sunnygleason:
> hmm I got an idea! We could use a trait to add a ::new() method how about that?

Yes I love that idea! Only problem is can't add from outside the crate?

Licenser:
If we control the trait we don't need to control the type :D and it compiles to a vld1 instruciton one way or the other so ¯\_(ツ)_/¯ doens't hurt to do the `as *const` dance in there - we take the hit complexity but it's nice for the user.

Licenser:
ahhh damn we can't have a trait with different arities for functions but we could pass in the slice, that's mildly better I think.

sunnygleason:
crazily enough I think mem::transmute made the most sense 😂 
right now, I'm up for whatever -- this is close enough

src/neon/generator.rs
sunnygleason: rock on, we can save that for another time

src/neon/stage1.rs
sunnygleason: I'm up for whatever, just be aware of my plan to adopt all the types so they can be made more friendly  :)

sunnygleason:
interesting looks like they do compile down to the same: https://rust.godbolt.org/z/0ORCzh

image

Licenser:
:D yay! I think the ::new is a lot nicer.

sunnygleason:
sorry I left this on the wrong PR, oops!

@Licenser did you find the bug? I think I might have it:

simd-lite/simdjson-rs@8875cff#diff-e3cc34fc9e09211fc01983167aaf86d7R47

(signed versus unsigned subtraction in utf8check, we had an error in the previous intrinsics)

PS. ignore all the intrinsics changes, that's just me bisecting the problem

Licenser:
Wow, well done! I think that did it!!! Once this all works we should give it a sweep to see if we can simplify some of those intriniscs but most importantly it seems to actually work :D good bye `std::arch`!

sunnygleason:
@Licenser this looks good to me! Are there big changes you'd want to make before releasing the simd-lite crate,are we getting close?

Licenser:
I really really would like to see if we can get rid of the copied code in simd-lite before releasing it wherever possible. And were not we need to make sure it's properly attributed and includes the respective copyrights.

sunnygleason:
Merging, beautiful work @Licenser !

PR #52, opened 2019-10-03
Fix more clippy errros and add docs

Licenser:


sunnygleason:
wow, this is a lot of work & documentation, very nice!

Licenser:
tagged as ` v0.1.21` and published :

PR #58, opened 2019-10-03
More tests

Licenser:
Add tests for code that isn't properly covered.

### DIFF
--- a/src/value/borrowed.rs
+++ b/src/value/borrowed.rs
@@ -284,6 +284,7 @@ impl<'de> BorrowDeserializer<'de> {
 
 #[cfg(test)]
 mod test {
+    #![allow(clippy::cognitive_complexity)]
     use super::*;
 
     #[test]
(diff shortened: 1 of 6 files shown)


sunnygleason:
A couple "interesting" things I'm observing:

* internal macros are being marked as untouched `stage2.rs`
* the `unsafe` leading line of an `unsafe` block is being marked as uncovered
* constants appear to be uncovered in at least a few cases as well...

Licenser:
Ja, I saw some of the same. In addition declaration without initialisation isn't covered either. It seems like rust code coverage is not all there yet, perhaps we should aim for a lower number then 90% given how much code is counted as not covered.

Licenser:
I added a ticket for a few of those: https://github.com/xd009642/tarpaulin/issues/262 if I missed any feel free to add them

sunnygleason:
@Licenser oh nice, that's awesome!

I didn't look into it that much -- do the macros we're using need the very tight/deep scope? Was wondering if they could be top-level decls...

Licenser:
Nah they need to exist in the function as they capture scpoe.

src/value/borrowed.rs
```diff
@@ -284,6 +284,7 @@ impl<'de> BorrowDeserializer<'de> {
 
 #[cfg(test)]
 mod test {
+    #![allow(clippy::cognitive_complexity)]
```
sunnygleason: 😂 🍺

Licenser: Yap sadly clippy doesn't understand context -.-

PR #63, opened 2019-10-08
Allow comparing refs

Licenser:
We lost the capability to compare &'s in the last cleanup lets add it back in a nicer way


Licenser:
merging since it's just api convenience

PR #64, opened 2019-10-08
Improve serde related tasks

Licenser:


src/serde/value/owned/de.rs
```diff
@@ -316,6 +316,7 @@ impl<'de> de::Deserializer<'de> for BorrowedCowStrDeserializer<'de> {
     where
         V: de::Visitor<'de>,
     {
+        dbg!();
```
sunnygleason: intentional? 😕

src/serde/value/owned/de.rs
```diff
@@ -275,7 +275,7 @@ impl<'de> serde::Deserializer<'de> for MapKeyDeserializer<'de> {
     {
         self.key
             .into_deserializer()
-            .deserialize_enum(name, variants, visitor)
+            .deserialize_enum(dbg!(name), dbg!(variants), visitor)
```
sunnygleason: intentional? 😕

.github/workflows/validate.yaml
```diff
@@ -62,6 +63,7 @@ jobs:
       - name: run tarpaulin
         env:
           RUSTFLAGS: '-C target-cpu=native'
+          PROPTEST_CASES: 512
```
sunnygleason: :+1:

src/serde/value/owned/de.rs
Licenser: nope just me being stupid :D good catch thanks!

src/serde/value/owned/de.rs
Licenser: same as below 😂!

.github/workflows/validate.yaml
Licenser: moatr porp's are always better ;)

Licenser:
note: not done yet still testing to be done on the serde parts

Licenser:
Okay we're now well over 80% and I can't see serde any more :joy: so I'm calling this a day.

Licenser:
added https://github.com/simd-lite/simdjson-rs/issues/68 as a followup

src/serde/value/owned/se.rs
```diff
@@ -657,4 +657,124 @@ impl serde::ser::SerializeStructVariant for SerializeStructVariant {
...
+mod test {
+    use crate::serde::from_slice;
+    /*
+    use crate::{
+        owned::to_value, owned::Object, owned::Value, to_borrowed_value, to_owned_value,
+        Deserializer,
+    };
+    use halfbrown::HashMap;
+    use proptest::prelude::*;
+    */
```
sunnygleason: curious, why are we keeping this code around? not a blocker!

src/serde/value/owned/se.rs
```diff
@@ -657,4 +657,124 @@ impl serde::ser::SerializeStructVariant for SerializeStructVariant {
...
+    enum Enum {
+        Opt1,
+        Opt2,
+    }
+    impl std::default::Default for Enum {
+        fn default() -> Self {
+            Self::Opt1
+        }
+    }
+    */
```
sunnygleason: ditto - curious, why are we keeping this code around? not a blocker!

PR #69, opened 2019-10-15
Adding KnownKey

Licenser:
Adding a feature for accessing known keys faster.

### DIFF
--- a/src/known_key.rs
+++ b/src/known_key.rs
@@ -15,6 +15,7 @@ pub struct KnownKey<'key> {
 /// Error for known keys
 #[derive(Debug, PartialEq, Clone)]
 pub enum Error {
+    /// The target passed wasn't an object
     NotAnObject(ValueType),
 }
 impl fmt::Display for Error {
--- a/src/known_key.rs
+++ b/src/known_key.rs
@@ -1,10 +1,10 @@
-use crate::{BorrowedValue, ValueTrait, ValueType};
+use crate::{ValueTrait, ValueType};
 use halfbrown::RawEntryMut;
-use std::borrow::Cow;
+use std::borrow::{Borrow, Cow};
 use std::fmt;
 use std::hash::{BuildHasher, Hash, Hasher};
 
-/// Well known key that can be looked up in a `BorrowedValue` faster.
+/// Well known key that can be looked up in a `Value` faster.
 /// It achives this by memorizing the hash.
 #[derive(Debug, Clone, PartialEq)]
 pub struct KnownKey<'key> {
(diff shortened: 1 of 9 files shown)


sunnygleason:
@Licenser this looks interesting -- does it just impact serde usage, because it's serialization/deserialization of a type with a more performant trait?

Licenser:
It doesn't affect serde compatibility or serialisation no. The effect is for BorrowedValue.get() (and I got to add OwnedValue.get() (performance improvement when using KnownKey). And deserialisation (performance decrease when using the `known-key` feature)

src/known_key.rs
```diff
@@ -0,0 +1,245 @@
...
+        }))
+    }
+}
+#[cfg(test)]
+mod tests {
+    #![allow(clippy::unnecessary_operation, clippy::non_ascii_literal)]
+    use super::*;
+    use crate::borrowed::*;
+    use crate::ValueTrait;
+
```
sunnygleason: Maybe an end-to-end test somewhere (perhaps with a doc comment with usage somewhere in the API?) would make things a bit clearer? These tests avoid the JSON processing APIs (targeted & good for a unit test, my brain was looking for an illustrative integration test(s)? I dunno...)

sunnygleason:
This seems very neat!

* I was looking for an illustration doc or integration test to "demo" usage
* could we add a benchmark case to show it works?

If you want to add a TODO comment/issue to the benchmarks in the right place, I'd be happy to pick that up!

Licenser:
codecov also seems to be broken :/

src/known_key.rs
```diff
@@ -0,0 +1,321 @@
...
+    ///
+    /// ```rust
+    /// use simd_json::*;
+    /// let object: BorrowedValue = json!({
+    ///   "answer": 42,
+    ///   "key": 7
+    /// }).into();
+    /// let known_key = KnownKey::from("answer");
+    /// assert_eq!(known_key.lookup(&object).unwrap(), &42);
+    /// ```
```
sunnygleason: this is awesome!

sunnygleason:
@Licenser this is very neat. So it's like precompilation of hash values on open objects (hashes) with known keys. Thank you for the implementation, and especially the docs, I think those turned out very nicely indeed!

src/known_key.rs
Licenser: added!

Licenser:
Thanks! It is a bit of a specialised case you really have the need for to use it but it can give some significant performance boosts. At the app I'm working on in $job there were some benchmarks where this went from 140 MB/s to 150MB/s throughput on this which is a nice speedup :D

sunnygleason:
> Thanks! It is a bit of a specialised case you really have the need for to use it but it can give some significant performance boosts. At the app I'm working on in $job there were some benchmarks where this went from 140 MB/s to 150MB/s throughput on this which is a nice speedup :D

That's interesting, I wonder if the field names add up to a lot of wasted copying as well. Could this case be further optimized to use `string -> int` indexes into a fixed-size `vec`, omitting fields?

Licenser:
Nah it's because we've build a language to do ETL on data, which happens to be JSON(esque) so JSONPath like expressions `this.that.here.there` do heavily rely on lookups in Objects. But since the script is only compiled ones but executed multiple times pre-paying for hashing and memorising it makes a difference.

Licenser:
released as v0.1.25 :)

sunnygleason:
sounds delightful! 😀

PR #71, opened 2019-10-23
0.2 work

Licenser:
This is a work branch for the 0.2 release where we can introduce breaking changes for things we do not like in 0.1.

So far:

- [x] Add U64 type
- [x] Remove deprecated functions
- [x] Box objects to reduce enum size
- [x] Add object and array access to value trait for convenience 
- [x] Arch alignment


Licenser:
@sunnygleason anything you can think of API wise we should change?

Licenser:
Either works for me. We can change everything other then the outside facing API as much as we want :D.

The 0.2 branch so far changes 2 things: it removes some functions that were bad and marked as deprecated in 0.1 already and adds the U64 as a natively supported type.

Since users will never notice how the parsing code works under the hood it's not constrained by the same restrictions :D.

Even adding 128 bit support (since we probably would want to feature flag that for perf reasons?)  would not change the API just add to it (which reminds me, no harm to add as_u128 and as_i128 to the value trait :D

sunnygleason:
how are we doing on benchmarks lately? I wonder, are there API changes or alternate entry points we could consider to increase parsing speed relative to native simdjson?

Licenser:
We don't have many benchmarks when it comes to the API, and that is a interesting topic since it is where simd-json.rs deviates a lot from simdjson.

There are 4 parts of the 'lifetime' of data:

1) decoding / parsing
2) reading / evaluating
3) editing / mutating
4) encoding / rendering

upstream simdjson is extremely good at 1) but makes 2) harder and probably slower and 3) nearly impossible. 4) I suspect will be faster as well since mutations do not happen.

A part of the speedup in 1) is by sacrificing 2) and 3) - instead of creating a DOM (like we do) it uses a tape, which is hands down faster - no allocation, no calculation of hashes, a unfragmented memory area.

However when reading that sacrifices a lot of convenience, you can't "take the `4th` element of an array" or "get the key `foo` from a object" all let alone do so in a nested way. That all has to be hand rolled, for objects every key/value pair has to be checked until the right key is found, for an array each element has to be passed to find the `n`th one.

Mutation (3)) with that model is simply not possible without first creating some kind of mutable structure out of it (which either means parsing into an object (which we have from serde support) or creating a DOM (the same way we do on parsing.

simdjson, as a low level library lets you pick your battles and if for example you only ever read, is significantly faster. simd-json.rs is more opinionated towards end to end workflow and convenience. That said I think on with parsing to objects we are not significantly slower

sunnygleason:
@Licenser wow, this is an enormous set of changes. Is there anything we can do to merge them in stages, or should I just swallow the big pill 💊 and review all at once?

sunnygleason:
ok only about 800 lines left to review! will come back to it soon. this is looking nice, great work! 🍻

Licenser:
you can look at the topic based PR's:
* https://github.com/simd-lite/simdjson-rs/pull/73
* https://github.com/simd-lite/simdjson-rs/pull/74

I create a pr towards the 0.2 branch for every change, just merge them every few days so they don't block as they're all dependant on each other.

sunnygleason:
gagh! answering a PR with more PR's, run away! 🏃 💨

Licenser:
they're merged already :P don't need to worry about them you just asked for smaller chunks, those where the chunks that make up this branch.

src/serde.rs
```diff
@@ -192,10 +164,11 @@ impl TryInto<serde_json::Value> for OwnedValue {
     fn try_into(self) -> ConvertResult<serde_json::Value> {
         use serde_json::Value;
         Ok(match self {
-            Self::Null => Value::Null,
-            Self::Bool(b) => Value::Bool(b),
-            Self::I64(n) => Value::Number(n.into()),
-            Self::F64(n) => {
+            Self::Static(StaticNode::Null) => Value::Null,
```
sunnygleason: 🎉

src/serde.rs
```diff
@@ -259,10 +223,11 @@ impl<'value> TryInto<serde_json::Value> for BorrowedValue<'value> {
     fn try_into(self) -> ConvertResult<serde_json::Value> {
         use serde_json::Value;
         Ok(match self {
-            BorrowedValue::Null => Value::Null,
-            BorrowedValue::Bool(b) => Value::Bool(b),
-            BorrowedValue::I64(n) => Value::Number(n.into()),
-            BorrowedValue::F64(n) => {
+            BorrowedValue::Static(StaticNode::Null) => Value::Null,
```
sunnygleason: 🎉

src/value.rs
```diff
@@ -89,6 +166,62 @@ pub trait ValueTrait:
...
+            .ok_or(AccessError::NotAnObject)
+            .map(|o| o.remove(k))
+    }
+
+    /// Tries to push to this `Value` as an `Array.
+    /// Will return an `AccessError::NotAnArray` if called
+    /// on a `Value` that isn't an `Array` - otherwise will
+    /// behave the same as `Vec::push`
+    #[inline]
+    fn push<V>(&mut self, v: V) -> std::result::Result<(), AccessError>
```
sunnygleason: nifty!

src/value.rs
```diff
@@ -151,69 +284,82 @@ pub trait ValueTrait:
...
+    #[inline]
     fn is_u128(&self) -> bool {
         self.as_u128().is_some()
     }
 
     /// Tries to represent the value as an u64
-    fn as_u64(&self) -> Option<u64> {
-        self.as_i64().and_then(|u| u.try_into().ok())
-    }
+    fn as_u64(&self) -> Option<u64>;
```
sunnygleason: where did the body go, is it overridden? 😕

src/value/borrowed.rs
```diff
@@ -22,28 +20,24 @@ pub type Object<'v> = HashMap<Cow<'v, str>, Value<'v>>;
...
 pub enum Value<'v> {
-    /// null
-    Null,
-    /// boolean type
-    Bool(bool),
-    /// float type
-    F64(f64),
-    /// integer type
-    I64(i64),
+    /// Static values
```
sunnygleason: 🎉

src/value/borrowed.rs
```diff
@@ -22,28 +20,24 @@ pub type Object<'v> = HashMap<Cow<'v, str>, Value<'v>>;
...
-    BorrowDeserializer::from_deserializer(de).parse()
+    match Deserializer::from_slice(s) {
+        Ok(de) => Ok(BorrowDeserializer::from_deserializer(de).parse()),
+        Err(e) => Err(e),
+    }
 }
 
 /// Borrowed JSON-DOM Value, consider using the `ValueTrait`
-/// to access it'scontent
+/// to access it's content
```
sunnygleason: the apostrophe in the comment can go, `its` vs `it's` ... yay English!

src/value/borrowed.rs
```diff
@@ -91,221 +80,251 @@ impl<'v> Value<'v> {
...
     fn value_type(&self) -> ValueType {
         match self {
-            Value::Null => ValueType::Null,
-            Value::Bool(_) => ValueType::Bool,
-            Value::F64(_) => ValueType::F64,
-            Value::I64(_) => ValueType::I64,
-            Value::String(_) => ValueType::String,
-            Value::Array(_) => ValueType::Array,
-            Value::Object(_) => ValueType::Object,
+            Self::Static(StaticNode::Null) => ValueType::Null,
```
sunnygleason: 🎉

src/value/borrowed.rs
```diff
@@ -91,221 +80,251 @@ impl<'v> Value<'v> {
 impl<'v> ValueTrait for Value<'v> {
     type Key = Cow<'v, str>;
 
+    #[inline]
```
sunnygleason: (same question about the more verbose `inline`)

src/value/borrowed.rs
```diff
@@ -91,221 +80,251 @@ impl<'v> Value<'v> {
...
     fn as_object_mut(&mut self) -> Option<&mut HashMap<Self::Key, Self>> {
         match self {
-            Value::Object(m) => Some(m),
+            Self::Object(m) => Some(m),
             _ => None,
         }
     }
 }
 
+#[cfg_attr(tarpaulin, skip)]
```
sunnygleason: interesting, why `skip`?

src/value/owned.rs
```diff
@@ -32,116 +32,129 @@ pub fn to_value(s: &mut [u8]) -> Result<Value> {
...
 pub enum Value {
-    /// null
-    Null,
-    /// boolean type
-    Bool(bool),
-    /// float type
-    F64(f64),
-    /// integer type
-    I64(i64),
+    /// Static values
```
sunnygleason: 🎉

examples/perf.rs
```diff
@@ -2,8 +2,8 @@ use std::env;
 
 #[cfg(feature = "perf")]
 mod int {
-    const ROUNDS: usize = 10000;
-    const WARMUP: usize = 2000;
+    const ROUNDS: usize = 2000;
```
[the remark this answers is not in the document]
Licenser: The benchmarks simply took too long, with the number of perf cases - it closed to 1h for a perf run which was just too long to get reasonable results

src/avx2/deser.rs
```diff
@@ -12,20 +12,26 @@ pub use crate::Result;
 
 impl<'de> Deserializer<'de> {
     // Allow it to keep in sync with upstream
-    #[allow(clippy::if_not_else)]
+    #[allow(clippy::if_not_else, mutable_transmutes, clippy::transmute_ptr_to_ptr)]
     #[cfg_attr(not(feature = "no-inline"), inline(always))]
-    pub fn parse_str_(&mut self) -> Result<&'de str> {
+    pub fn parse_str_<'invoke>(
```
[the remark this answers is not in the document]
Licenser: Lifetime names can be picked like variable names. Sadly the rust default u I s to pick one letter but I tend to use descriptive names to help with understanding what they do. It seems to work;) yes it’s borrowed for only the  invocation of the function

Cargo.toml
```diff
@@ -49,21 +49,28 @@ harness = false
...
+# uses safe slice access ([]) instead of get_unsafe
+# **for debugging**
+safe = []
 # also bench serde in the benchmarks
 bench-serde = []
 # use branch hints - requires nightly :(
 hints = [] # requires nightly
 # for perf testing, used by the example
 perf = ["perfcnt", "getopts", "colored"]
-known-key = [ "halfbrown/fxhash" ]
```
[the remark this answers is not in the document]
Licenser: Merge oopsy, good catch thanks!

.github/workflows/validate.yaml
```diff
@@ -40,14 +40,14 @@ jobs:
         run: bash <(curl https://raw.githubusercontent.com/xd009642/tarpaulin/master/travis-install.sh)
       - name: run tarpaulin
         env:
-          RUSTFLAGS: '-C target-cpu=native -C target-feature=-avx2'
+          RUSTFLAGS: "-C target-cpu=native -C target-feature=-avx2"
           PROPTEST_CASES: 512
-        run: cargo tarpaulin -v --features no-inline --out Xml -- --test-threads=1 && cp cobertura.xml ssl.xml
```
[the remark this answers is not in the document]
Licenser: Yup 0.9.1 fixed the threads issue:)

Cargo.toml
```diff
@@ -49,21 +49,28 @@ harness = false
...
-# Support for ARM NEON SIMD
-neon = ["simd-lite"]
+# used for enabeling known keys in favour of a slower
+# hasher that is not protected against hash collision
+# attacks
+known-key = [ "halfbrown/fxhash" ]
 # use 8 number at once parsing strategy
 swar-number-parsing = []
-# serde compatibility
+# serde compatibilitya
```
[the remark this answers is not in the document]
Licenser: Ja! Thanks

src/error.rs
```diff
@@ -95,23 +93,23 @@ pub struct Error {
 }
 
 impl Error {
-    pub(crate) fn new(structural: usize, index: usize, character: char, error: ErrorType) -> Self {
```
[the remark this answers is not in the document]
Licenser: It was a bit very helpful number, people should not need to bother about which structural index it is - it is implementation detail

Licenser:
> The only other things I was wondering about were:
> 
> * would it be useful to specify what version of simdjson upstream this is tracking?
Yes! But we'd need to figure out what it is :joy:
> * arch alignments to template-ize code, would that still be useful? (while we're changing everything 😝 )
Yes! either towards 0.1 or 0.2 I'd love if we'd get there! But as always, provided it's not a perf hit - sometimes generic code is not as fast as specific code, but we got our benchmarks and perf tests to decide on that ;)
> 
> In any case, this looks really nice, the benchmark results sound outstanding!

Cargo.toml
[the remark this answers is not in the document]
Licenser: fixed

Cargo.toml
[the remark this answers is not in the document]
Licenser: Oh no, it just moved up, I re-ordered the features a bit to put the more relevant ones to the top and stuff like perf or bench related things that no one would use in production towards the end :)

src/value/borrowed.rs
Licenser: fixed!

src/value/borrowed.rs
Licenser: that actually could be simplified away :) since value_type is implemented for StaticNode. Good catch!

src/value/borrowed.rs
Licenser: It's just a Display implementation that dispatches into the underlying rust types, so if that would go wrong we've waaayyyy bigger problems - aka not worth putting testing effort into it.

src/value/owned.rs
Licenser: :D

src/value/borrowed.rs
Licenser: I'm always unsure, I tend to give the `#[inline]` hint for tiny functions to push rustc/llvm more towards considering to inline them (as it's replacing 1 command with 1 command) and `#[inline(always)]` for things that we know we want inlined (based on upstream), and the rest I do not touch either and trust that rustc/llvm knows better then me :P

Licenser:
@sunnygleason lets do a quick planning on the 0.2 release, I think getting the "this is following simdjson vXXX in is a good goal to have. but we'd need to pick a version and then go over it to see how we track. Would you want to get the arch alignment changes into 0.2 as well? They kind of fit to that theme. Is there anything I'm missing for 0.2 we'd want

sunnygleason:
@Licenser I like that Value work you're doing. I'm doing an investigation of turning SimdInput into a struct/trait with associated functions to reduce/standardize code in `stage1`. Should have something (hopefully interesting) in the next couple days. Warm up the benchmarks!  😄  🍻

Licenser:
Sweet!

Licenser:
I'm good with merging this, @sunnygleason what do you think?

sunnygleason:
@Licenser yes, I think we've got it! If you think it could wait one more day, I was thinking of doing the `utf8check` and "upstream tracking version" updates early tomorrow morning (ET)...

Licenser:
👍 absolutely, as long as there is stuff to be done we can wait :) put a not down when you're ready to go .- I want to avoid the everyone is waiting for someone else problem ;) so as soon as you're ready, I'm ready.

sunnygleason:
@Licenser I think that's all the tricks I have for now, release anytime!  😁

Licenser:
Perf:

image


## master
```
======= simd_json ======== parse|stringify ===== parse|stringify ====
data/canada.json         410 MB/s   370 MB/s   620 MB/s   280 MB/s
data/citm_catalog.json  1060 MB/s   320 MB/s  1410 MB/s   380 MB/s
data/twitter.json        900 MB/s   410 MB/s   900 MB/s   370 MB/s
data/log.json           1080 MB/s   720 MB/s  1080 MB/s   720 MB/s
```

## this
```
======= simd_json ======== parse|stringify ===== parse|stringify ====
data/canada.json         530 MB/s   430 MB/s   710 MB/s   320 MB/s
data/citm_catalog.json  1130 MB/s   480 MB/s  1550 MB/s   610 MB/s
data/twitter.json        920 MB/s   600 MB/s  1000 MB/s   650 MB/s
data/log.json           1080 MB/s  1080 MB/s  1080 MB/s  1080 MB/s
```

Licenser:
0.2 release
* upstream tracking version
* generalisation of utf8parsing
* generalisation of stage1
* api update
* perf improvements
* tape parsing

PR #72, opened 2019-10-24
Add ::from for Object

Licenser:
Add From<Object> for values for convenience (trivial change)


Licenser:
merging since it has no wider impact.

PR #76, opened 2019-11-05
Try pointers instead of str for strings

Licenser:
This PR would use a pointer and 32bit length instead of a &str on the tape - that way the size of an tape node is reduced to 16 byte and string creation is handled after the fact.

Turns out despite the space savings this is degrading performance - I'll open and close this PR for tracking purposes.


Licenser:
closing since it's bad for perf

sunnygleason:
💔 😢

Licenser:
oops

PR #79, opened 2019-11-19
Split traits into read, mutate and build

Licenser:
This is a pure API change and should have no performance impact. The goal here is to split out different aspects for the value trait to allow implementing only parts of it. Since 0.2 is a declared breaking API it feels like the right time to do.

This will allow us to do things like make non mutable Value variants that have better performance or even serialise directly to objects.


Licenser:
This is I think what's needed for the new Value Trait's - ready for review / merge

PR #83, opened 2019-12-01
Compare <= for alloc tests

Licenser:
simple test fixes to deal with inconsistencies. The inconsistency are likely based on an additional allocation to be introduced when we're too close to a page boundary.


Licenser:
Merging since it's just housekeeping

Issue #101, opened 2020-02-07
Use with serde_derive


Timmmm:
Is it possible to use this crate with `serde_derive` to speed up parsing? E.g. currently I have

```
#[derive(Serialize, Deserialize, Debug)]
pub struct Foo {
  pub a: u64,
  pub b: Vec<u64>,
  pub c: Option<u64>,
}
...

  let file = File::open(filename)?;
  let reader = BufReader::new(file);
  let result: Foo = serde_json::from_reader(reader)?
```

Is there a way to do something like this?

```
  let result: Foo = simd_json::from_reader(reader)?
```

Licenser:
Hi Timmmm!

There is `from_slice(&mut [u8])` that will do what you are looking for for the most part. We currently do not have a 'from_reader' interface since simdjson depends on being able to go over the entire data in a go.

I would have to look into if it is possible to wrap a from_reader around it but I'm slightly unsure given we do use the input value for references (this might not affect it, but I don't want to give false hope).

How far does from_slice get you?

https://docs.rs/simd-json/0.2.3/simd_json/serde/fn.from_slice.html

Timmmm:
Ah yes this seems to work!

```
  let file = File::open(filename)?;
  let mut data = Vec::new();
  file.read_to_end(&mut data)?;
  let v: serde_json::Value = simd_json::serde::from_slice(&mut data)?;
  let result = Foo::deserialize(v)?;
```

Might be worth adding to the readme because it wasn't obvious.

Unfortunately this was waaay slower than just using the default Serde implementation for me. Serde is ~4s to read a 185 MB file, this was 9s. Oh well!

Licenser:
Are you compiling your code using native CPU detection settings?

Licenser:
That is either copying the .cargo from this project (or making your own with the flag) or using `RUSTFLAGS="-C target-cpu=native"`

Timmmm:
Yeah I put that in my `.cargo/config` and am compiling with `--release`. I guess converting to `serde_json::Value` and then again to my structs is slower than just reading into the structs directly?

Licenser:
Oh, yes sorry I totally missed that part! I kind of thought you're reading to a Vec then deserialising it.

That explains the speed I'll try to explain:

First you 'pay' for DOM deserialization - which is significantly slower than deserializing a struct since instead of knowing already how the memory layout looks like it will try to piece it together on the fly as the data comes in.

[...]

PR #102, opened 2020-02-07
Add from_reader method for serde integration

Licenser:
#101 brought this up, we're missing the from_reader method, so here it is.


Timmmm:
That was fast!

Licenser:
We try :)

Licenser:
@sunnygleason any chance to give this a quick look?

Issue #34, opened 2019-07-31
0.1.17 does not compile


dtolnay:
The 0.1.17 release is failing to build in Travis, when 0.1.16 used to work. Any idea what may be going on here?
https://travis-ci.org/serde-rs/json-benchmark/jobs/565755287

<details>
<summary>error messages</summary>

```console
error[E0425]: cannot find value `SIMDJSON_PADDING` in this scope
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/simd-json-0.1.17/src/stage2.rs:256:48
    |
256 |                 let mut copy = vec![0u8; len + SIMDJSON_PADDING];
    |                                                ^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `SIMDJSON_PADDING` in this scope
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/simd-json-0.1.17/src/stage2.rs:271:48
    |
271 |                 let mut copy = vec![0u8; len + SIMDJSON_PADDING];
    |                                                ^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `SIMDJSON_PADDING` in this scope
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/simd-json-0.1.17/src/stage2.rs:286:48
    |
286 |                 let mut copy = vec![0u8; len + SIMDJSON_PADDING];
    |                                                ^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `SIMDJSON_PADDING` in this scope
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/simd-json-0.1.17/src/lib.rs:142:79
    |
142 |         let needs_relocation = (buf_start + input.len()) % page_size::get() < SIMDJSON_PADDING;
    |                                                                               ^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `SIMDJSON_PADDING` in this scope
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/simd-json-0.1.17/src/lib.rs:145:62
    |
145 |             let mut data: Vec<u8> = Vec::with_capacity(len + SIMDJSON_PADDING);
    |                                                              ^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `SIMDJSON_PADDING` in this scope
[...]

PR #103, opened 2020-02-07
Switch to not compiling with a non SIMD capable cpu settings

Licenser:
#101 shows that allowing fallback is a tricky beast. It leads to users having a bad experience, that shouldn't happen while not compiling is annoying having a slow system for non-obvious reason from a library that is about speed is worse. 

This PR disables it and creates an error when compiled w/o see42, avx2 or NEON.  It can be re-enabled with the `allow-non-simd` flag.

solves #34

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,8 @@ swar-number-parsing = []
 serde_impl = [ "serde", "serde_json", "halfbrown/serde" ]
 # Support for ARM NEON SIMD
 neon = ["simd-lite"]
+# Allow fallback to non simd CPUs
+allow-non-simd = []
 # for testing allocations
 alloc = ["alloc_counter"]
 # don't inline code - used for debugging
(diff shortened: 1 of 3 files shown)


sunnygleason:
@Licenser can you clarify the platform where this happened?

Licenser:
Basically when a CPU doesn't support SSE4.2. Turns out the issue was a different one, but it can just happen when target-cpu isn't set during compilation.

It's a super easy mistake to make, especially when simd-json isn't included from the top level.

I think an error that explains what is wrong is better than something suddenly being slow.

sunnygleason:
@Licenser that is interesting! I think your approach makes sense. I wanted to also bring up runtime dispatch -- will do that in a separate issue. Cheers! 🍻

Licenser:
#102 has the clippy fixes in it I'll rebase this once #102 is merged then it should be all green :)

Issue #61, opened 2019-10-07
align number parsing with upstream


Licenser:
Upstream has done some changes to number parsing we should resync

sunnygleason:
Speaking of syncing, was interested in your thoughts, see #62 🏎

Licenser:
this was done in #253

PR #127, opened 2020-04-24
Implement number parsing from simdjson v0.3.1

ijl:
This includes accurate and consistent float parsing. Float parsing
is generally handled by the fast path but falls back to lexical-core.

This passes number tests from JSONTestSuite except
n_multidigit_number_then_00.json because of the trailing padding.


ijl:
I think this closes #61.

Licenser:
heya this fails to build with:
```
error[E0554]: `#![feature]` may not be used on the stable release channel
   --> /home/heinz/.cargo/registry/src/github.com-1ecc6299db9ec823/lexical-core-0.7.4/src/lib.rs:133:35
    |
133 | #![cfg_attr(not(feature = "std"), feature(core_intrinsics))]
    |                                   ^^^^^^^^^^^^^^^^^^^^^^^^
```

ijl:
Ok, fixed with the last commit, I was on nightly.

Licenser:
## This

```
                                DOM                  STRUCT
======= serde_json ======= parse|stringify ===== parse|stringify ====
data/canada.json         270 MB/s   310 MB/s   480 MB/s   220 MB/s
data/citm_catalog.json   430 MB/s   240 MB/s   950 MB/s   300 MB/s
data/twitter.json        300 MB/s   410 MB/s   600 MB/s   470 MB/s
data/log.json            360 MB/s   540 MB/s   720 MB/s   720 MB/s

======= simd_json ======== parse|stringify ===== parse|stringify ====
data/canada.json         380 MB/s   410 MB/s   530 MB/s   310 MB/s
data/citm_catalog.json  1010 MB/s   390 MB/s  1480 MB/s  1240 MB/s
data/twitter.json        910 MB/s   590 MB/s  1060 MB/s  1530 MB/s
data/log.json           1080 MB/s  1080 MB/s  1080 MB/s  2170 MB/s

===== simd_json_tape ===== parse|stringify ===== parse|stringify ====
data/canada.json         570 MB/s                    
data/citm_catalog.json  1850 MB/s                    
data/twitter.json       1650 MB/s                    
data/log.json           2170 MB/s                    
```

## Master
```
                                DOM                  STRUCT
======= serde_json ======= parse|stringify ===== parse|stringify ====
data/canada.json         270 MB/s   320 MB/s   520 MB/s   220 MB/s
data/citm_catalog.json   430 MB/s   240 MB/s   930 MB/s   300 MB/s
data/twitter.json        300 MB/s   420 MB/s   600 MB/s   470 MB/s
data/log.json            430 MB/s   720 MB/s   720 MB/s   720 MB/s

======= simd_json ======== parse|stringify ===== parse|stringify ====
data/canada.json         460 MB/s   420 MB/s   690 MB/s   310 MB/s
data/citm_catalog.json  1020 MB/s   400 MB/s  1490 MB/s  1260 MB/s
data/twitter.json        910 MB/s   590 MB/s  1060 MB/s  1540 MB/s
data/log.json           1080 MB/s  1080 MB/s  1080 MB/s  2170 MB/s

===== simd_json_tape ===== parse|stringify ===== parse|stringify ====
data/canada.json         770 MB/s                    
data/citm_catalog.json  1870 MB/s                    
data/twitter.json       1640 MB/s                    
data/log.json           2170 MB/s                    
```

Licenser:
It looks like this PR does significantly reduce performance for number parsing, any idea why?

ijl:
I'm not famliar with the previous implementation, but it seems reasonable that performance is worse for accurate parsing. I can't use the old implementation because it won't roundtrip common floats such as 0.3 nevermind harder cases, so I think performance is kind of irrelevant until it's correct.

Licenser:
> so I think performance is kind of irrelevant until it's correct.
I generally agree, however a 20% performance degradation is something worth investigating. simdjson (upstream) uses https://github.com/lemire/fast_double_parser which si fairly fast (and correct). master uses simdjsons old parsing and from what I remember on the release notes for simdjson 3 the new number parsing so I'm a bit hesitant to just assume there is no way around it :) especially.

Basic rule of thumb:

If it's faster and better / corrected: merge no questions asked
if it's slightly slower and better / corrected: merge and cry privately
if it's significantly slower and better / corrected: perhaps feature flag it so users can choose?


This looks like a 3) but I kind of hope that with a bit of time it can be a 2).

So I'll spend some time looking over the code perhaps there is a 'easy' fix to move it closer to original speed :). If not we'll feature flag the old code so people who don't need the correctness in float pasring can opt out of it, sounds good?

Licenser:
So it turns out the canada.json always hits the slow part of f64_from_parts that is where the 'pain' comes from here is a look into perf:
image

Licenser:
So bottom line I can't find anything obvious. Upstream simdjson got marginally slower on parsing canda.json however nowhere close to 20%. I don't want to hold off this PR so simple solution, can you copy the current `numberparse.rs` into a new file and put it behind feature flag such as `approx-float-parsing` that way people can pick and it's minimal work?

Licenser:
the clippy complaints test shouldn't be your problem :tm: we got to fix them so I'll run tests locally to verify and we bypass CI on that.

Licenser:
I've been looking into making the approximate parsing optional but I noticed that the current PR would remove support for 128-bit numbers which is problematic.

I can share my current state of swapping algorithms if you're interested.

Licenser:
I got the feature flagged code here: https://github.com/simd-lite/simd-json/tree/ijl-parse-number it also cleans up all clippy warnings, feel free to use that as a starting point :)

Licenser:
@ijl ping?

Licenser:
oops brnach naming confusion didn't meant to close it, can you re-open it vs `main` ideally with the fixes from jil-parse-number

Issue #151, opened 2020-10-03
Ability to reuse Deserializers to avoid allocations


CJP10:
`simd_json::to_borrowed_value(bytes)` seems to allocate two buffers, `string_buffer` and `buffer`. I think it makes sense to have these be arguments, similar to `Deserializer::from_slice_with_buffer` to allow reuse of all buffers.

Licenser:
A very good observation! You are absolutely right :)

CJP10:
@Licenser I'll start working on a PR.

I do have a question, what is trying to do?
```
let align = buffer
            .as_slice()
            .as_ptr()
            .align_offset(SIMDJSON_PADDING / 2);
        unsafe {
            buffer
                .as_mut_slice()
                .get_unchecked_mut(align..align + len)
                .clone_from_slice(input);
            *(buffer.get_unchecked_mut(len + align)) = 0;
            buffer.set_len(len + align);
        };
```

Licenser:
This attempts to align the writing position in the buffer to the size of a SIMD register so we don't have unaligned memory reads. We then add a 0 at the end to ensure that we don't have any JSONesque characters trailing. i.e. otherwise we could run in the situation where the last character is a 1 and then the number 123 could become 1231 (rust strings are non terminated this way we can safe on checking lenght there)

CJP10:
How do you feel about having a wrapper to enforce algimnet?

```
pub struct AlignedBuf {
    ptr: NonNull<u8>,
    cap: usize,
}

impl AlignedBuf {
    pub fn new(capacity: usize) -> Self {
        let ptr = Layout::from_size_align(capacity, SIMDJSON_PADDING / 2).unwrap();
        Self {
            ptr: ptr.into(),
            cap: capacity,
        }
    }

    fn into_vec(self) -> Vec<u8> {
        unsafe { Vec::from_raw_parts(self.ptr.as_ptr(), 0, self.cap) }
    }
}
```

CJP10:
@Licenser made a rough draft of my idea in #152.

PR #152, opened 2020-10-03
Ability to reuse Deserializers to avoid allocations

CJP10:
closes https://github.com/simd-lite/simd-json/issues/151


Licenser:
Looks good 👍 especially the change in aligned buffer looks very neat :D. I think the only reason it errors is that the `fn error` currently only gets included for serde removing https://github.com/CJP10/simd-json/blob/reuse-deserializers/src/lib.rs#L397 should fix that.

Licenser (approved):
Thank you this looks great :+1:

Licenser:
released as `v0.3.21`

PR #163, opened 2021-01-21
Add to_value for borrowed value

Licenser:
add serde `to_value` and Serializer for `BorrowedValue<'static>`


mfelsche:
It seems like rustc is not happy about some allow/forbid flags conflicting between platform specific and unspecific code? :(

PR #164, opened 2021-01-21
Add parse functions in parallel with to_value

Licenser:
add `parse` functions that mirror `to_value` for borrowed and owned value


mfelsche (commented):
LGTM, except the clippy complaints :)

Issue #166, opened 2021-02-06
Bad UTF-8 accepted due to incorrect Utf8Check::is_incomplete() implementation


Count-Count:
The `Utf8Check::is_incomplete()` implementation is broken for sse42 and avx2 (did not check neon). Saturatingly substracting 255 of every byte yields 0 for every byte so a multi-byte character at the end of a SIMD block is not checked when the next block is processed.

Testcase for avx2 (256-bit blocks):
```rust
fn main() {
    let bad_utf8_bstr = b"\"123456789012345678901234567890\xC2134567890\"";
    assert!(std::str::from_utf8(bad_utf8_bstr).is_err()); // from_utf8() errors out as expected
    let mut d = bad_utf8_bstr.to_vec();
    assert!(simd_json::to_borrowed_value(&mut d).is_err()); // supposed to error because of bad UTF-8
}
```

PR #169, opened 2021-02-11
Fix utf8 is incomplete fn on avx and sse42 (issue #166)

Count-Count:
For upstream implementation see https://github.com/simdjson/simdjson/blob/af4db55e669cf0c9d4ea281685ae51039c382d76/src/generic/stage1/utf8_lookup4_algorithm.h#L111


Licenser:
Hyea :) 
thanks for the contribution! Very good catch.

It seems like the issues are related to ab update in calippy on rust 1.50.0, nothing you should be responsible to update so I'll :+1: this without the clippy tests passing.

Licenser (approved):
:+1:

PR #174, opened 2021-03-25
Fix owned value u64 serializer to require u64

darach:
Note - quick change via web edit feature - but I suspect this is ( one of ) the causes of as_u64 failing property based tests in tremor


Licenser (approved):
Nice :+1:

PR #176, opened 2021-03-29
Add json_typed macros for borrowed json literal support

darach:
Signed-off-by: Darach Ennis <darach@gmail.com>

### DIFF
--- a/src/serde/value/borrowed/se.rs
+++ b/src/serde/value/borrowed/se.rs
@@ -1,8 +1,12 @@
-use crate::value::borrowed::Value;
+use super::to_value;
+use crate::value::borrowed::{Object, Value};
 use crate::StaticNode;
+use crate::{stry, Error, ErrorType, Result};
 use serde_ext::ser::{
     self, Serialize, SerializeMap as SerializeMapTrait, SerializeSeq as SerializeSeqTrait,
 };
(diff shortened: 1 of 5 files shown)


src/serde/value/borrowed/se.rs
```diff
@@ -529,9 +529,7 @@ impl<'a> serde_ext::Serializer for MapKeySerializer<'a> {
 
     #[inline]
     fn serialize_str(self, value: &str) -> Result<Self::Ok> {
-        // TODO: This is ugly
-        //let s = value.to_owned();
-        Ok(value)
+        Ok(unsafe { std::mem::transmute(value) }) // TODO find a better way to serialize &str
```
Licenser, line 532: We need to find a way around this, this will give a segfault if str isn't living as long as `'se`

Licenser: We could try to return a `Cow` here instead of a `&str` that way we can just use `Cow::from(str.to_string())`

Licenser: Or even `String` perhaps?

Licenser (approved):
Nice thank you so much! This will be super useful :D

Licenser:
released as `0.4.1` :+1:

PR #178, opened 2021-04-09
Remove unnecessary dbg call

Gelbpunkt:
Spams console output - maybe left over from an old PR.


Licenser:
Oh darn it, sorry about that good catch and thank you for fixing it!

Licenser:
released as `0.4.2`

PR #179, opened 2021-04-12
Remove another dbg macro call

Gelbpunkt:
Hi there, sorry for this second PR. :flushed: 
I oversaw another `dbg!()` in the other PR which was also there. The last PR was really just copy-from-stacktrace-and-remove-in-fork, but this is according to the search the last call to the debug macro left in the codebase. Maybe I should've looked a bit closer before submitting that one.


Licenser:
Don't be sorry after all I released it with two debug calls in it :D

Thank you a bunch for finding them! I'm super curious what you're using simd-json for? Anything you can share?

PR #181, opened 2021-04-18
Fix serialization of numeric map keys

as-com:
Previously, serializing a struct like:

```rust
#[derive(Clone, Debug, PartialEq, serde::Serialize)]
struct Foo {
    pub bar: HashMap<i32, i32>,
}
```

would result in incorrect JSON output (no quotes around the keys). This PR adds a MapKeySerializer that corrects this behavior by wrapping numeric keys in quotes to more closely match the behavior of serde_json, and also adds a test case.

### DIFF
--- a/src/serde/se.rs
+++ b/src/serde/se.rs
@@ -223,10 +223,11 @@ where
 
         if *first {
             *first = false;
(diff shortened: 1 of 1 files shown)


Licenser:
Another one :) you're on fire! Thank you again!

PR #182, opened 2021-04-18
Allow Value to accept numeric map keys as serialization target

as-com:
Previously, calling `simd_json::serde::to_{owned,borrowed}_value` on a struct like this:
```rust
#[derive(Clone, Debug, PartialEq, serde::Serialize)]
struct Foo {
    pub bar: HashMap<i32, i32>,
}
```

would result in an error. This PR makes BorrowedValue and OwnedValue more closely behave like `serde_json::Value` by also accepting numeric map keys.

### DIFF
--- a/src/serde/value/borrowed/se.rs
+++ b/src/serde/value/borrowed/se.rs
@@ -471,63 +471,52 @@ impl<'se> serde_ext::Serializer for MapKeySerializer<'se> {
         Err(key_must_be_a_string())
     }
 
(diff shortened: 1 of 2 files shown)


Licenser:
Nice one :D thank you! This is a solid improvement, I'll re-run the CI as the issues seem to be unrelated.

Licenser (approved):

Licenser:
@as-com, again thanks for all the contributions :) they're bundled up and released as `v0.4.4`

PR #183, opened 2021-04-18
Fix incorrect output when serializing unknown length zero-length maps

as-com:
Previously, serializing a struct like this:
```rust
#[derive(Debug, Serialize)]
struct Foo {
    #[serde(flatten)]
    bar: HashMap<u32, bool>,
}
```

would produce incorrect output (a single opening curly brace) when the HashMap is empty. This is due to incorrect logic surrounding calls to `serialize_map` with an unknown length, which occurs when using the flatten feature. This PR fixes this bug.


Licenser:
Thank you! Good catch :)

PR #187, opened 2021-05-02
Add KnownKey::map* methods for lookups on Value::Objects inner maps.

mfelsche:
This is handy if you already know you have a `Value::Object(map)` in your hands, in which case we can spare a match on the enum variant.

This is a backport of what we did in https://github.com/tremor-rs/tremor-runtime/pull/969


Licenser (approved):
Thank you! 👍

PR #192, opened 2021-05-13
Use rustup beta, which fixes the spurious OOM issue on Windows 

hkratz:
See https://github.com/rust-lang/rustup/issues/2759


Licenser (approved):
Makes sense to me! Thank you

hkratz:
That was most likely the reason for this build failure: https://github.com/simd-lite/simd-json/runs/2576604318?check_suite_focus=true

Licenser:
good catch!

Issue #200, opened 2021-07-15
Illegal Instruction (core dumped)


superblaubeere27:
Important to note:
- According to `cat /proc/cpuinfo`, my cpu supports SSE4.2.
- I have the `.cargo/config` of this repository in place
- I am using rustc 1.53.0

After compiling the program, it crashes with the message `Illegal Instruction (core dumped)`.

How can I fix this?

Licenser:
There are a few possibilities here, usually the issue means that the compilation was done with features that aren't supported by the CPU, I got a few ideas on that :)

1) Do you compile and run the code on the same machine? If yes then you can use `rustflags = ["-C", "target-cpu=native"]` in the `.cargo/config`
2) If you do not, you might have a wrong setting in the `.cargo/config` if the CPU you run it on has SSE4.2 the setting that **should** :tm: work is `rustflags = ["-C", "target-feature=+sse4.2"]`

If neither of that works could you share your `/proc/cpuinfo` and `.cargo/config`? :)

superblaubeere27:
Thank you for your quick reply.

I compiled the program on the same machine I ran it on and I had the `rustflags = ["-C", "target-cpu=native"] ` in `.cargo/config`.

Unfortunately also the setting you suggested (`+sse4.2`) did not work and the program still crashes due to the CPU not recognizing an instruction.

Here are the requsted files:

### `.cargo/config` (Right now):
```TOML
[build]
rustflags = ["-C", "target-feature=+sse4.2"]
```
### `.cargo/config` (When making this issue):
```TOML
[build]
rustflags = ["-C", "target-feature=native"]
```
### `car /proc/cpuinfo`:
```
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 26
model name      : Intel(R) Xeon(R) CPU           E5530  @ 2.40GHz
stepping        : 5
microcode       : 0x1d
cpu MHz         : 2533.562
cache size      : 8192 KB
physical id     : 0
siblings        : 8
core id         : 0
cpu cores       : 4
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 11
wp              : yes
[...]

PR #201, opened 2021-07-16
Made the compatiblity check fail on absense of the `pclmulqdq`-feature

superblaubeere27:
Preventing the library from producing invalid code on some machines (See #200)


Licenser (approved):
Awesome :D thank you!

PR #203, opened 2021-07-19
Added fallback function for CPUs that don't support pclmulqdq instructions

### LINKED ISSUE #202: Remove hard dependency on `pclmulqdq` support
Followup to #200.

We should have a fallback for replacing `_mm_clmulepi64_si128` on CPUs that do not support `pclmulqdq`  this could be done by making a fallback implementation for `compute_quote_mask`

superblaubeere27:
I was unable to create a faster function with SIMD instructions, primarily because x86 has no 64-bit vector multiply instruction. 

On my machine, the fallback function makes the parsing ~10% slower

Fixes #202

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -13,6 +13,7 @@ jobs:
         rustflags:
           - '-C target-cpu=native'
           - '-C target-cpu=native -C target-feature=-avx2'
+          - '-C target-cpu=native -C target-feature=-avx2,-pclmulqdq'
         features:
           - ''
           - '--features known-key'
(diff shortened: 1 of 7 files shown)


Licenser:
Awesome :D code looks great! Thank you so much for the quick fix of the `pclmulqdq` issue :). It will just need a `cargo fmt --all` to make clippy happy I think otherwise :+1: ready to merge and I'll release a `0.4.8` with it

Licenser:
Looks nice :+1: it seems to need a rebase to solve conflicts before we can merge it :)

superblaubeere27:
> 
> 
> Looks nice 👍 it seems to need a rebase to solve conflicts before we can merge it :)

I can't find any merge conflicts. Or would that fix the Coverage QA failing?

Licenser:
No the coverage is a different topic, but GH claims there is  a conflict with main and prohibits merging:

grafik

Licenser (approved):
looks great :D thank you!

PR #206, opened 2021-10-12
pre alloc maps in the `json!` macro

### LINKED ISSUE #180: improve the json! macro to pre-allocate maps
Right now, in `json!` maps are build using `Object::new()` then inserting keys into them repeatedly, leading to possible extra locations. It would be nice to do this with `with_capacity` there are some options for counting sub elements here: 
https://danielkeep.github.io/tlborm/book/blk-counting.html#repetition-with-replacement

and 

here: https://docs.rs/halfbrown/0.1.11/src/halfbrown/macros.rs.html#46-62

fisherdarling:
This PR adds support to the `json!` macro to pre allocate maps with the number of inserted keys.

Rather than immediately inserting when a valid entry was found in the TT muncher, we append to a stack of valid entries. Once we run out of entries to parse, we count the entries and then insert them all at once. We use this count to create a new `Object::with_capacity(count)` before inserting the keys.

closes #180

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -455,7 +455,7 @@ macro_rules! json_internal_owned {
     // TT muncher for parsing the inside of an array [...]. Produces a vec![...]
     // of the elements.
     //
-    // Must be invoked as: json_internal!(@array [] $($tt)*)
+    // Must be invoked as: json_internal_owned!(@array [] $($tt)*)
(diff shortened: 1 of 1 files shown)


Licenser:
Awesome, thank you I'm quite busy today so I might not get around to review it right away but I'll do it this week :)

PR #207, opened 2021-10-29
Add more tests to serde feature

dvermd:
Add some tests as part of #57

Here's a first PR with some working tests.

I'm still working on map coverage and I think it will come in another PR

Issue #57, opened 2019-10-03
Improve testing


Licenser:
We still have a rather low test coverage on parts of the code (my fault -.-) now that we can measure it we should aim to improve it. 90%+ sounds like a good target

sunnygleason:
Yay! You found my not-so-subtle reason for adding coverage 😂

Licenser:
To be fair the coverage of the critical path is quite good, >90% even with the buggy coverage. The tests missing are more for 'boring/unimporant' functions and will just exist to push the number and make it look nicer.

sunnygleason:
👍 , my 100% coverage club membership expired a couple years ago... 😂

Licenser:
I have lots of opinions on the topic 😂

sunnygleason:
getting there! the big thing that jumps out at me here are:

* [ ] serde
* [ ] stage2.rs

Wondering if there's a way to get the macros out of the function definition in `stage2.rs` ...

Licenser:
Ja serde is the big thing left.

dvermd:
I think there's some dead code in the serde part of this crate. Can you help identify if this is correct ?

1. In `serde/value/owned/se.rs`, I didn't manage to trigger any function of `MapKeySerializer` other than `serialize_str`. I think this is because, internally, the Object type of a Value is a `HashMap<String, Value>` and thus, only String keys will be serialized
2. In `serde/value/owned/de.rs`, the `ValueVisitor::visit_i/u 8/16/32` should be called from the `parse_integer` from serde_json but it only ever returns a i64, a u64 or a f64 as of https://github.com/serde-rs/json/blob/master/src/de.rs#L380.
3. Same for the borrowed values

Licenser:
We have a lot of tests now

PR #209, opened 2021-10-29
Improve coverage #2

dvermd:
This PR is based on the PR #207 

- Add the ability to deserialize HashMaps from Tape.
- Add some more tests as part of #57
- Do not run coverage on test code


dvermd:
It seems that the code coverage is tagged as decreased because it uses the number of lines as metrics. With the `--ignore-tests` option of tarpaulin, there are less lines that are to be covered. Looking at Coveralls report, we can that the `relevant` column decreased just like the `covered` one.

If you want to review this change in its own PR to better see the coverage improvements of the other commits, let me know and I'll split the PR.

Licenser:
Sorry for the late reply :) this looks great I'll update the branch and see if that solves the clippy complaining

Licenser (approved):

Licenser:
Seems update was not enough :) can you address the clippy issues and then  we can merge this :D

dvermd:
I had a hard time finding the clippy warnings the previous commits added. So I fixed most of the clippy warnings (most of them by `clippy --fix`).
The only 2 remaining are from `json!` and `json_typed!` macro that swallow the error on inserting a (key, value) pair to a Value. I don't know how you want to handle this one but I think hiding the error this way will be misleading at some point for a user.

Licenser:
:+1: awesome, I think leaving that out for now is fine. the `let _ = ...insert` in the macros isn't swallowing errors but rather swallowing the return value as we already know that the object is an object, and the content is a value since it's created by a macro :)

PR #210, opened 2021-11-01
Improve coverage3

dvermd:
This PR is based on the PR #209 and adds more tests for #57 

I added 2 methods to the API to be able to trigger some of the deserializing code.
If these methods should not go in the API, the corresponding code can be removed


Licenser (approved):
This looks great, thank you so much, you are the best!

Licenser:
This looks great :) I think only thing missing is that in `stage2.rs` it should be `crate::serde::from_slice` for the tests to pass?

Licenser (approved):
Awesome, this is incredible work :D! Thank you so  much!

PR #214, opened 2021-12-10
Fix alloc issues with alligned buf

### LINKED ISSUE #213: Soundness bug in AlignedBuf implementation
@gkorland recently tried to add `simd_json` to the comparison suite for my `ijson` crate, but it failed the tests due to an allocation failure.

Upon investigation, it appears that there's a bug in the `AlignedBuf` struct from this crate.

https://github.com/simd-lite/simd-json/blob/ae48dc4ee34c8cc698c73ac599bcfdd648a53cd4/src/lib.rs#L679

The rust allocator API requires that the same layout be passed to both `alloc` and `dealloc`. However, this implementation uses a much larger alignment when allocating, but an alignment of 1 when deallocating. To be correct, `AlignedBuf` needs to override both allocation, deallocation *and any potential reallocation* of the underlying memory.

For that reason, I would suggest not wrapping `Vec` at all, since any `Vec` methods which might resize the backing store would be invalid to call.

Licenser:
This removes the vector for the AllignedBuf and instead keeps the pointer and a custom drop implementation.

This should fix #213


mfelsche (approved):
Looks legit!

PR #219, opened 2022-03-01
feat(arch): add support for wasm32 with simd128 feature

### LINKED ISSUE #217: WASM support
Whilst WASM doesn't expose the native CPU instructions, it does have its own SIMD intrinsics (see https://doc.rust-lang.org/core/arch/wasm32/index.html#simd).

I had a quick look at porting this crate to WASM this evening and it seems to be feasible. I've not had a chance to test it in the slightest, but it builds. (https://github.com/sam-kirby/simd-json/tree/wasm)

I'll try including it in a yew project I'm working on tomorrow and see if it a) works and b) offers any performance improvements.

_Originally posted by @sam-kirby in https://github.com/simd-lite/simd-json/issues/161#issuecomment-1043640140_

sam-kirby:
This adds support for two targets: 
1) `wasm32-unknown-unknown` - the current go to for running Rust on web browsers. The standard library is incredibly limited here and running tests is difficult.
2) `wasm32-wasi` - currently most useful for running tests, but may become more desirable in the future due to powerful sandboxing features offered by wasm runtimes on server.

In theory, any future wasm32 targets will be supported without any modification - e.g. `wasm32-web`. I've not made any effort to support `wasm64` yet, though enabling this in the future should be a simple matter of changing some cfg macros.

Some key changes and issues:
- proptest is incompatible with wasm targets as it relies on being able to fork, which is not possible. As a result, this dev dependency has to be gated and tests using it turned off on wasm targets. Due to `unused_imports` being a denied lint, certain imports in tests had to be broken up to allow compilation of tests on wasm32.
- testing requires wasmtime; a make rule has been added to make this as easy as possible. This could be substituted for any other wasm/wasi runtime.
- simdutf8 is currently a git dependency; support for wasm has not yet been included in a release of this crate
- value-trait also needed modification to support this, and I will open a separate PR there. Due to breaking changes this PR will target value-trait's 0.2.1 tag, not the current head.
- benchmarking doesn't work as libc ends up in the dependency graph somehow - I've not looked into this yet.
- whilst comparing the implementation for wasm to the neon version I noticed that the `vtstq_u8` intrinsic is now part of the standard library and so removed the temporary implementation. Do let me know if you'd like this reverted.
- wasm currently lacks any kind of runtime feature detection (though this should be possible in the future through [exception handling](https://github.com/WebAssembly/exception-handling/blob/master/proposals/exception-handling/Exceptions.md)), and simd instructions may be disabled (though in most cases they seem to be enabled by default). This may cause issues when considering the `allow-non-simd` feature.

closes #217

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,2 +1,8 @@
 [build]
 rustflags = ["-C", "target-cpu=native"]
+
+[target.wasm32-unknown-unknown]
+rustflags = ["-C", "target-feature=+simd128"]
+
+[target.wasm32-wasi]
+rustflags = ["-C", "target-feature=+simd128"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,8 +14,8 @@ rust-version = "1.49"
 [dependencies]
 beef = { version = "0.5", optional = true }
 halfbrown = "0.1"
-value-trait = { version = "0.2.1" }
-simdutf8 = { version = "0.1.3", features = ["public_imp", "aarch64_neon"] }
+value-trait = "0.2.10"
+simdutf8 = { version = "0.1.4", features = ["public_imp", "aarch64_neon"] }
 
 # serde compatibilty
 serde = { version = "1", features = ["derive"], optional = true}
@@ -35,6 +35,8 @@ criterion = { version = "0.3", optional = true }
 [dev-dependencies]
 float-cmp = "0.9"
 getopts = "0.2"
+
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 proptest = "1.0"
 
 [lib]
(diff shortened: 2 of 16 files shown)


Licenser:
Heya that is some great stuff!

I merged the value-trait PR it's tagged as 0.2.10 so that temporary dependency can go away, once the other one PR is merged I think we can merge this one too :) 

I'm quite excited, WASM support is really great! Thank you!

Licenser:
I suspect the remaining overwrite is because we're waiting for a new release of simdutf8?

sam-kirby:
That is correct, yes

hkratz:
I will do a new release of simdutf8 shortly.

hkratz:
simdutf8 v0.1.4 with WASM support has been released.

Licenser (approved):
Awesome, thank you!

Licenser:
Weeeeh!

PR #233, opened 2022-07-04
Revert "Improve the error message emitted when a SIMD-incompatible CPU is used"

Licenser:
Heya @scrabsha I've to reverts simd-lite/simd-json#232 for now as now all the doc tests fail :(m sorry


scrabsha:
Hi there!

To be honest I don't really understand why it failed with `compile_error`, as both methods should work the same.

Anyways, sorry for the inconvenience and thank you for quickly fixing my mistakes :D

PR #252, opened 2022-10-11
Fix bench file for criterion v0.4.0

philss:
This is needed in order to execute the benchmarks. They were based in the previous version of criterion.


Licenser (approved):

Licenser:
nice thanks!

PR #259, opened 2022-12-01
Add value-no-dup-keys flag

Licenser:
Add a feature flag for strict (non UB) parsing of duplicate keys in objects for Value structs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,8 +49,8 @@ harness = false
 [features]
 default = ["swar-number-parsing", "serde_impl"]
 
-# Forces the `owned::Value` and `borrowed::Value` to check for duplicated keys. This comes at a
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -49,6 +49,11 @@ harness = false
 [features]
 default = ["swar-number-parsing", "serde_impl"]
 
+# Forces the `owned::Value` and `borrowed::Value` to check for duplicated keys. This comes at a
```
mfelsche, line 52:
```suggestion
# Forces the `owned::Value` and `borrowed::Value` to deduplicate duplicated keys by letting consecutive keys overwrite previous ones. This comes at a
```

mfelsche (approved):
Do we have any benchmark comparisons for this change?

PR #262, opened 2022-12-15
Variant access tests and cleanup

Licenser:
cleanup serializer and deserializer implementations

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -68,7 +68,7 @@ pub enum ErrorType {
     Serde(String),
     /// Generic syntax error
     Syntax,
-    /// Training characters
+    /// Trailing data
     TrailingData,
     /// Unexpected character
     UnexpectedCharacter,
--- a/src/error.rs
+++ b/src/error.rs
@@ -197,6 +197,6 @@ mod test {
     #[test]
     fn fmt() {
         let e = Error::generic(ErrorType::InternalError);
-        assert_eq!(format!("{}", e), "InternalError at character 0");
+        assert_eq!(e.to_string(), "InternalError at character 0");
     }
 }
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,15 +1,17 @@
 use std::fmt;
 
+use value_trait::ValueType;
+
 /// Error types encountered while parsing
 #[derive(Debug)]
 pub enum ErrorType {
+    /// A specific type was expected but another one encountered.
+    Unexpected(Option<ValueType>, Option<ValueType>),
     /// Simd-json only supports inputs of up to
     /// 4GB in size.
     InputTooLarge,
     /// The key of a map isn't a string
     BadKeyType,
-    /// The data ended early
-    EarlyEnd,
     /// Expected an array
     ExpectedArray,
     /// Expected a `,` in an array
@@ -67,11 +69,9 @@ pub enum ErrorType {
     /// Generic syntax error
     Syntax,
     /// Training characters
-    TrailingCharacters,
+    TrailingData,
     /// Unexpected character
     UnexpectedCharacter,
-    /// Unexpected end
-    UnexpectedEnd,
     /// Unterminated string
     UnterminatedString,
     /// Expected Array elements
@@ -99,7 +99,6 @@ impl PartialEq for ErrorType {
         match (self, other) {
             (Self::Io(_), Self::Io(_))
             | (Self::BadKeyType, Self::BadKeyType)
-            | (Self::EarlyEnd, Self::EarlyEnd)
             | (Self::ExpectedArray, Self::ExpectedArray)
             | (Self::ExpectedArrayComma, Self::ExpectedArrayComma)
             | (Self::ExpectedBoolean, Self::ExpectedBoolean)
@@ -127,9 +126,8 @@ impl PartialEq for ErrorType {
             | (Self::Parser, Self::Parser)
             | (Self::Eof, Self::Eof)
             | (Self::Syntax, Self::Syntax)
-            | (Self::TrailingCharacters, Self::TrailingCharacters)
+            | (Self::TrailingData, Self::TrailingData)
             | (Self::UnexpectedCharacter, Self::UnexpectedCharacter)
-            | (Self::UnexpectedEnd, Self::UnexpectedEnd)
             | (Self::UnterminatedString, Self::UnterminatedString)
             | (Self::ExpectedArrayContent, Self::ExpectedArrayContent)
             | (Self::ExpectedObjectContent, Self::ExpectedObjectContent)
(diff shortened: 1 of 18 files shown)


src/serde/value/borrowed/de.rs
```diff
@@ -141,32 +142,79 @@ impl<'de> de::Deserializer<'de> for Value<'de> {
     }
 }
 
-struct Array<'de, 'value: 'de>(std::slice::Iter<'de, Value<'value>>);
+struct Array<'de>(std::vec::IntoIter<Value<'de>>);
```
darach, line 145: Nice!

src/serde/value/borrowed/de.rs
```diff
@@ -537,8 +551,7 @@ impl<'de> VariantAccess<'de> for VariantDeserializer<'de> {
                 if v.is_empty() {
                     visitor.visit_unit()
                 } else {
-                    visitor.visit_seq(Array(v.iter()))
-                    // visit_array_ref(v, visitor)
+                    visitor.visit_seq(Array(v.into_iter()))
                 }
             }
             // FIXME
```
darach, line 557: Is this vestigial? Perhaps we can just remove the FIXME comment here

src/serde/value/borrowed/de.rs
```diff
@@ -564,8 +577,8 @@ impl<'de> VariantAccess<'de> for VariantDeserializer<'de> {
     {
         match self.value {
             Some(Value::Object(v)) => visitor.visit_map(ObjectAccess {
-                i: v.iter(),
-                v: &Value::Static(StaticNode::Null),
+                i: v.into_iter(),
+                v: Value::Static(StaticNode::Null),
             }),//visit_object(v, visitor),
             // FIXME
```
darach, line 583: Same. Remove comments that aren't adding value.

src/serde/value/borrowed/de.rs
```diff
@@ -784,10 +802,10 @@ impl<'de> VariantAccess<'de> for VariantRefDeserializer<'de> {
         V: Visitor<'de>,
     {
         match self.value {
-            Some(Value::Object(v)) => visitor.visit_map(ObjectAccess {
+            Some(Value::Object(v)) => visitor.visit_map(ObjectRefAccess {
                 i: v.iter(),
                 v: &Value::Static(StaticNode::Null),
-            }),//visit_object(v, visitor),
+            }),
             // FIXME
```
darach, line 809: Same. FIXME and commented out code not adding anything

src/serde/value/borrowed/de.rs
```diff
@@ -1048,4 +940,49 @@ mod test {
...
+            config: Option<borrowed::Value<'v>>,
+        }
+        impl<'v> serde::Deserialize<'v> for NameAndConfig<'v> {
+            fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+            where
+                D: serde::Deserializer<'v>,
+            {
+                #[derive(Deserialize)]
+                // This is ugly but it's needed for `serde::Deserialize` to not explode on lifetimes
+                #[serde(bound(deserialize = "'de: 'v, 'v: 'de"), untagged)]
```
darach, line 957: We'll forget the lifetime pinning hack in 2 weeks and will be puzzled in 6 months as to what this is doing and why
Also future rusts and/or serde's may be more ( or less ) strict here.
Perhaps add the compiler error when this bound annotation is omitted would suffice?

src/serde/value/owned/de.rs
```diff
@@ -842,7 +752,11 @@ impl<'de> VariantAccess<'de> for VariantRefDeserializer<'de> {
                 if v.is_empty() {
                     visitor.visit_unit()
                 } else {
-                     visit_array_ref(v, visitor)
+                    // This is somewhat nasty, but it's worth it since we can otherwise
+                    // avoid the clode by using into_iter instead of iter.
```
darach, line 756: `clode` -> `clone` otherwise `clode` is a good name for a project!

src/serde/value/owned/se.rs
```diff
@@ -12,24 +16,24 @@ impl Serialize for Value {
         S: ser::Serializer,
     {
         match self {
-            Self::Static(StaticNode::Bool(b)) => serializer.serialize_bool(*b),
-            Self::Static(StaticNode::Null) => serializer.serialize_unit(),
-            Self::Static(StaticNode::F64(f)) => serializer.serialize_f64(*f),
-            Self::Static(StaticNode::U64(i)) => serializer.serialize_u64(*i),
+            Value::Static(StaticNode::Null) => serializer.serialize_unit(),
```
darach, line 19: Naming. My brain trips on `serialize_unit` when I see it in the context of `null`. Nothing to resolve here though, not sure `serialize_null` is any better as we cannot distinguish `null` value from `null` reference and some uses may imply `undefined` because JSON. Perhaps its right my brain should be trip right here!

src/serde/value/borrowed/de.rs
Licenser: the `FIXME` here is for the error handling (or lack thereof)

src/serde/value/borrowed/de.rs
Licenser: same :sob:

src/serde/value/owned/de.rs
Licenser: outdated comment, I'll yank it completely (we're no longer cloning)

src/serde/value/owned/se.rs
Licenser: I agree but sadly not in our control, this naming is dictated by `serde` :(

src/serde/value/borrowed/de.rs
Licenser: same, same :sob:

perhaps I should just fix the fixmes ... I will do that tomorrow

src/serde/value/borrowed/de.rs
darach: 😭

src/serde/value/borrowed/de.rs
darach: 👍

src/serde/value/owned/de.rs
darach: Even better - whack the comment entirely!

src/error.rs
```diff
@@ -67,11 +69,9 @@ pub enum ErrorType {
     /// Generic syntax error
     Syntax,
     /// Training characters
```
Agares, line 71:
```suggestion
    /// Trailing data
```

darach (approved):
🚀

Agares (approved):

src/error.rs
Licenser: Good catch, thanks!

PR #276, opened 2023-04-17
Use dtolnay's rust-toolchain GH Action

philss:
This is an attempt to remove all the warnings related to deprecations in GitHub Actions.

The main reason for this change is because the Action-RS' toolchain is unmaintained since Nov 2020. This new GH Action now in use is maintained by one of the Rust Core Team members.


Licenser (approved):
Good catch, nice idea, I'm all for it 👍

PR #294, opened 2023-05-18
Fix neon implementation for cpu without feat_pmull

### LINKED ISSUE #293: PMULL feature is unavailable on some cpu
I couldn't run simd-json on raspberrypi (Illegal instruction), because it's missing PMULL arm feature.
vmull_p64 is used in https://github.com/simd-lite/simd-json/blob/7b8e0bd8d6b04124499ca0050203cd5614da6e17/src/neon/stage1.rs#L69 without feature check/alternative impl.

Mesteery:
alternative implementation taken from https://github.com/simdjson/simdjson/blob/master/include/simdjson/arm64/bitmask.h

I wonder if it would be better to just have the version of simdjson since it is stated that it is more powerful than pmull?

Closes #293

### DIFF
--- a/src/neon/stage1.rs
+++ b/src/neon/stage1.rs
@@ -66,20 +66,6 @@ impl SimdInput {
 
 impl Stage1Parse<int8x16_t> for SimdInput {
     #[cfg_attr(not(feature = "no-inline"), inline(always))]
-    #[cfg(target_feature = "pmull")]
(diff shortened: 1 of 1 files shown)


Licenser:
Actually reading the upstream source it seems that someone did the analysis that this is faster then pmul all together, lets replace the original implementation completely as you suggested in #293 then we'll merge this

Licenser (approved):
Awesome!

Issue #290, opened 2023-05-16
Consider using ahash with a seed


ritchie46:
Hi, I read the `known_key` optimization, using fxhash because the hash is stable.

A `RandomState::with_seeds` would also produce stability, but favor the faster hash function and better hash quality. I think this can improve performance.

We especially noticed in polars that performance of fxhash can drop of a cliff on some data patterns (factors of 2 for instance).  https://github.com/tkaitchuck/aHash/blob/master/compare/readme.md#fxhash

Any thoughts?

Licenser:
Hi,

I like the idea. That'd be a nice improvement, if nothing else, something we could put behind a feature flag to let people choose what they want.

One thing that brings up is how to pick a seed, it could be pre-randomized on every startup with something like a lazy static, or settable by the user in a global variable. Since you've a usecase right now, what would work best for you there?

ritchie46:
Assuming you don't want the `RandomState` in the public API. You can pick a fixed seed on startup with lazy static/ once cell. Or even simpler is a lazy OnceCell with a random initialized `RandomState` in it. The size of the `RandomState` is only 4 `u64`, so no problem keeping that around.

https://rust-lang.github.io/hashbrown/src/ahash/random_state.rs.html#88-189

Licenser:
This is quite a bit more complicated than anticipated, AHasher doesn't seem to have an interface that allows initializing it with either a given RandomSeed, nor one that allows initializing it with keys:

https://docs.rs/ahash/0.8.3/ahash/struct.AHasher.html

ritchie46:
No, it is the other way around. A `RandomState` produces a stable `AHasher`. 

`RandomState` implements the [BuildHasher](https://doc.rust-lang.org/nightly/core/hash/trait.BuildHasher.html) trait. So calling `RandomState::build_hasher()` returns an `AHasher`.

Here is the source to confirm: https://rust-lang.github.io/hashbrown/src/ahash/random_state.rs.html#198-233

Licenser:
I know, `RandomState` is the Builder and gets passed as a generic type to the HashMap. So we'd have to make a new struct I snarkily called it `NotSoRandomState` that acts as a `BuildHasher` and generates an `AHasher` the problem is that the hash builder has to call `AHasher::from_random_state` and that's a private function.

ritchie46:
Have you got a small snippet that shows the problem?

I believe we can store the Randomstate in a static and clone it when we need it to instantiate a HashMap. 

[...]

PR #295, opened 2023-05-22
Move to ahesh for known keys as suggested in #290

Licenser:
@ritchie46 want to take a look and see if / how this works for you?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,8 +18,12 @@ features = ["docsrs"]
 lexical-core = { version = "0.8", features = ["format"] }
 beef = { version = "0.5", optional = true }
 halfbrown = "0.2"
-value-trait = { version = "0.6" }
+value-trait = { version = "0.6.1" }
 simdutf8 = { version = "0.1.4", features = ["public_imp", "aarch64_neon"] }
+
+# ahash known key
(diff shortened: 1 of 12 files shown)


ritchie46:
Yes! I shall take a look! Thanks already.

Cargo.toml
```diff
@@ -18,9 +18,14 @@ features = ["docsrs"]
...
 beef = { version = "0.5", optional = true }
 halfbrown = "0.2"
-value-trait = { version = "0.6" }
+value-trait = { version = "0.6.1" }
 simdutf8 = { version = "0.1.4", features = ["public_imp", "aarch64_neon"] }
 
+# ahash known key
+once_cell = { version = "1.17", optional = true }
+ahash = { version = "0.8", optional = true }
+rand = { version = "0.8", optional = true }
```
ritchie46, line 27: Why is `rand` needed? Couldn't find any usage of it in this PR.

ritchie46 (commented):
I went through the whole PR and I think this would work. :+1: 

With the `ObjectHasher` it seems abstracted away nicely.

Licenser:
I'll try to release this as 0.10 today (as the changes in hasher might be breaking backwards compatibility)

PR #308, opened 2023-08-21
Use pointers instead of slices to avoid tripping the optimizer in nightly

Licenser:
Switch to using pointers for `input` in string parsing to avoid fighting the borrow checker and tripping up the optimizer in nightly.


mfelsche (approved):
Great fix!

PR #318, opened 2023-09-18
Fix docs build

mkrasnitski:
Right now, local builds of the docs for simd-json fail due to the "target-cpu" compilation error. This unfortunately also happens on docs.rs, although one would it expect it doesn't. This fixes it by changing the `cfg(docsrs)` to `cfg(doc)`, which should apply both to local doc builds as well as when publishing to docs.rs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,9 +11,6 @@ readme = "README.md"
 documentation = "https://docs.rs/simd-json"
 rust-version = "1.61"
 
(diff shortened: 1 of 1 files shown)


Licenser:
Heya,

sorry for the slow reply. I wanted to get #317 merged first as resolving conflicts this way around was easier :) :+1: 

this looks good, thanks a lot for the contribution! 🚀

PR #323, opened 2023-09-28
refactor tape::Node<'input>

PSeitz:
The two parameters in Object and Array are undocumented and not
understood without looking at the source code.
Refactor to be more explanatory and add docs

I'm not sure of the usefulness of `end`. It's not used currently and maybe can be removed?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -875,7 +875,7 @@ mod tests {
         let mut d = String::from("[]");
         let d = unsafe { d.as_bytes_mut() };
         let simd = Deserializer::from_slice(d).expect("");
-        assert_eq!(simd.tape[1], Node::Array { len: 0, end: 2 });
+        assert_eq!(simd.tape[1], Node::Array { len: 0, count: 0 });
     }
 
     #[test]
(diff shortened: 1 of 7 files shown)


Licenser:
I really like this change :) I've stumbled over this myself a few times and always told myself "I should fix that" then forgot 😂 so thanks!

One thought, should we move from recording the `end` (end offset on the tape) to `count` (tape elements that are included in the array/object)? I think the `count` would make a few things easier like creating a wrapper around the tap for iterating over it w/o turning it into a nested struct like simd_json (upstream) has.

PSeitz:
> One thought, should we move from recording the `end` (end offset on the tape) to `count` (tape elements that are included in the array/object)? I think the `count` would make a few things easier like creating a wrapper around the tap for iterating over it w/o turning it into a nested struct like simd_json (upstream) has.

Hmm I'm not sure how `count` would help with a wrapper to iterate, can you give an example?
Even `len` is not really needed for correctness, only for preallocation.

Licenser:
The [skip function](https://github.com/simd-lite/simd-json-derive/blob/main/src/lib.rs#L13) is a good example, we need to recursively go over the elements to skip the right number of nodes since we only have the total offset (`end`) not the number of elements that need to be skipped (`count`)

PSeitz:
> The [skip function](https://github.com/simd-lite/simd-json-derive/blob/main/src/lib.rs#L13) is a good example, we need to recursively go over the elements to skip the right number of nodes since we only have the total offset (`end`) not the number of elements that need to be skipped (`count`)

I updated the PR the replace `end` with `count`

Licenser (approved):
Nice :D,

great improvements!

PR #326, opened 2023-10-10
reuse structural_indexes

PSeitz:
structural_indexes allocations can be saved in some scenarios

### DIFF
--- a/benches/parse.rs
+++ b/benches/parse.rs
@@ -11,8 +11,7 @@ static ALLOC: jemallocator::Jemalloc = jemallocator::Jemalloc;
 use serde_json;
 
 use criterion::{criterion_group, BatchSize, Criterion, Throughput};
-use simd_json::AlignedBuf;
-use simd_json::SIMDJSON_PADDING;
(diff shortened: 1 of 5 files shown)


Licenser (approved):
I quite like this change 👍 great observation. Saving an allocation here can make a real difference for long-running applications!

How do you feel about creating a struct with the three-pointers that can be passed instead of having the three of them passed separately? To be fair, this is not something that this PR introduced, we already had two parameters, but with three, it feels like we're reaching the point where something like a parser-context (or something like that) could start to be good?

PSeitz:
> How do you feel about creating a struct with the three-pointers that can be passed instead of having the three of them passed separately? To be fair, this is not something that this PR introduced, we already had two parameters, but with three, it feels like we're reaching the point where something like a parser-context (or something like that) could start to be good?

Yes, I added a `Buffers` struct. The usage was already a little bit clunky.

Licenser (approved):

Licenser (approved):
I really like this 👍

PR #328, opened 2023-10-19
Native impls

Licenser:
This PR adds a native implementation that can act as a fallback for platforms that do not support any of the implemented architectures.

As a addition, it adds a `std::simd` implementation that would allow for native (but possibly not as optimized) simd on any platform that supports rust, but since this feature is a nightly feature, it is guarded behind the `portable` feature flag


src/portable/stage1.rs
```diff
@@ -0,0 +1,195 @@
+use std::{
+    ops::Shr,
+    simd::{prelude::*, ToBitMask},
+};
+
+macro_rules! low_nibble_mask {
```
darach, line 6: Needs a comment or it'll be magic to use in a few weeks!

Licenser: I'll move the code to the comment that bleongs to it :D, good call!

Licenser:
this requires https://github.com/rusticstuff/simdutf8/pull/83 to be merged and released to, we depend on simdutf8 for string validation

PR #339, opened 2023-10-27
Fix aarch64 build

zertosh:
Various recent refactors have left the project unbuildable on aarch64.
CI seems to be stuck for this arch, so there's been no recent signal
that things are broken. This is currently affecting the 0.13.0 release.

Test Plan:

```
$ cargo build --all --target aarch64-apple-darwin
$ cargo test --target aarch64-apple-darwin

$ cargo build --all --target x86_64-apple-darwin
$ cargo test --target x86_64-apple-darwin
```


Licenser:
Good catch, thanks I wonder why drone never failed, I've to investigate that. :+1:

Licenser (approved):

PR #350, opened 2023-10-29
Update README.md - fix grammatrical errors in README.

sonijogesh:
I have noticed some grammatical mistakes in **README** inside section "**runtime-detection**"

"**The reason being that rust **"  should be "**The reason is that rust **"

" **optimise as far to the **"  should be " **optimise as far as the **"

Screenshot-

Screenshot (111)


sonijogesh:
Please review my PR.

Licenser (approved):
Thanks :)

PR #363, opened 2024-01-10
Treat big ints as floats via feature flag "big-int-as-float"

richarddd:
Currently simd-json errors when a to large integer overflows while parsing.

This PR adds a feature flag `big-int-as-float` that handles cases when int overflows and interprets that as a float instead.
The downside is that you loose some precision but that's ok if you opt into it and how it currently works in most JavaScript based engines.

For example:
`JSON.parse("999999999999999999999999999999")` in Chrome (v8) returns 1e+30

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -129,6 +129,7 @@ bench-canada = []
 bench-citm_catalog = []
 bench-log = []
 bench-twitter = []
+big-int-as-float = []
 
 [[example]]
 name = "perf"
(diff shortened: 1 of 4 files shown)


Licenser:
Good idea 👍 I like it; makes sense that in some places, not failing on icky input is more important than being 100% correct; the only thing could you add a section to the [README.md where the other feature flags are listed](https://github.com/simd-lite/simd-json#runtime-detection)  that explains the flag and the restrictions?

richarddd:
> Good idea 👍 I like it; makes sense that in some places, not failing on icky input is more important than being 100% correct; the only thing could you add a section to the [README.md where the other feature flags are listed](https://github.com/simd-lite/simd-json#runtime-detection) that explains the flag and the restrictions?

Done! 🥇

Licenser:
released as `v0.13.7`

PR #364, opened 2024-01-19
Fix compatibility on unsupported CPU architectures

liushuyu:
This pull request fixes the issue where the `simd-json` fallback code path isn't working for unsupported CPU architectures.

`#[allow(dead_code)]` is added for the `handle_unicode_codepoint` function since this function is not used when building the fallback logic.

An alternative method would be changing the `Cargo.toml` file to gate `runtime-detection` feature behind `x86_64` and `x86` architectures.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -482,7 +482,10 @@ impl<'de> Deserializer<'de> {
         }
     }
     #[cfg(not(any(
-        feature = "runtime-detection",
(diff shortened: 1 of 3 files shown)


Licenser (approved):
I like it, thanks!

Licenser:
released as `0.13.8` thanks a lot :)

Issue #365, opened 2024-01-24
simd_json::from_reader (and from_slice) end in error when trying to deserialize to a serde_json::RawValue


bassmanitram:
The error is "invalid type: newtype struct, expected any valid JSON value".

The use case is demonstrated by [this GIST](https://gist.github.com/bassmanitram/fc58b50915e3266004a808431e34a050). Changing the version of `lambda_http` to 0.8 makes the use case work.

The difference is that the AWS code is now trying to deserialize to a RawValue where it previously serialized to a `serde::__private::de::Content`. Obviously the new code is far cleaner, but it doesn't work with the simd deserializer.

Any help would be appreciated.

(Equivalent issue in AWS world - https://github.com/awslabs/aws-lambda-rust-runtime/issues/792)

Licenser:
I'm currently on vacation so I won't have time to look deeply into it for a while but at a first glance I think RawValue is the issue. Unrelated to that this is going to be fairly slow so not much benefit from the decode is to be had.

The slowness comes from that the bytes first goes to a nested structure (RawValue) this is expensive, then is translated to a struct using the serde traits (expensive again).

If you were out for performance I'd say use to_tape, try to introspect the content for the type, then turn that in a sturct. That would be no overhead to directly translating it to a struct.

bassmanitram:
Hey - have a good vacation. Yeah, I too spotted the slowness of the approach and am talking to the AWS guys about it - thanks for the `tape` tip!

I have narrowed the problem down to this basic code that fails (yes it is RawValue):

```
	let mut payload3 = "{}".to_string();
	let payload3 = unsafe {payload3.as_bytes_mut()};
	//let mut de = serde_json::Deserializer::from_slice(payload3);
	let mut de = simd_json::Deserializer::from_slice(payload3).unwrap();
	let raw_value: Box<RawValue> = Box::deserialize(&mut de).expect("Boxed RawValue");
```

As above, it fails. Uncomment serde_json and comment out simd_json and it works.

bassmanitram:
[...]

PR #366, opened 2024-01-28
Allow the tape to be used in a standard serde::Deserialize

bassmanitram:
Issue #365 

* The new `from_tape` method on the Deserializer allows a tape to be extracted from a Deserializer via `into_tape` and then be used to rebuild a Deserializer. It is marked as `unsafe` because no structure verification is performed - it is up to the caller to make sure that the tape is correct (e.g. by extracting from a Deserializer, performing read-only operations, then rebuilding a Deserializer).

At the moment, however, nothing I try is able to allow access to that from within a standard serde::Deserializer.

* Deserialize in a tap: really complex to implement
* Extension Trait - specialization not allowed
* ?


Licenser (approved):
Looks good :) I was about to suggest that on the issue when you submitted the PR 👍  thanks a lot!

Licenser:
Oh I just saw, it'll need a `cargo fmt` before passing the CI

bassmanitram:
K, I'm on it

bassmanitram:
Hold of merging for now - the larger goal is to figure out how to take the deserializer in a serde::de::Deserialize implementation's deserialize method, get the Tape from it, do my "type determinant" searches, then turn the Tape back into the deserializer and pass it onto the discovered types. Obviously this PR enables that last bit, but it's getting the Tape from the "anonymous" deserializer that has me a bit stuck at the moment.

Licenser:
I think  a way of achieving this would be changing the implementation from `Deserializer for Deserializer` (yay naming ...) to `Deserilizer for &[Node]` or `Deserializer for Vec<Node>` that way the same implementation would work for `Deserializer` and `Tape` since both have the vec/slice?

bassmanitram:
Since Deserialize is foreign and so are Vec and any slice notation, we need a local Trait to cover them... thinking

bassmanitram:
Ok, that's totally untested at the moment, but compiles and passes clippy - I'm now going to see if, at least in my use case, it works.

bassmanitram:
You live and learn eh - can't "specialize", it seems - the default implementation is always invoked!

bassmanitram:
OK, lets try a deserializer for Node ... which of course means I have to start hacking another of your libraries - value_trait :)

bassmanitram:
Nope - that's fantastically complex too, it seems! This is becoming soul destroying!

Licenser:
Just a few thought experiments would any of the following work:

1) move the implementation of `serde::Deserializer` from `simd_json::Deserializer` to `simd_json::Tape`
2) change the according functions that use the `deserializer.deserialize()` call to call `deserializer.into_tape().deseriaze()`

Or:

adding a `as_tape(&self)  -> &[Node]` function and use that to inspect the type

bassmanitram:
OK, I leave this here - serde::de::DeserializeSeed "solves" the issue, but where I'm trying to implement it has serde_json so ingrained that swapping over to simd_json (with necessary "unsafe" stuff too) is just too onerous at the moment.

So the PR with only from_tape is ready, and some kind of solution to RawValue would be appreciated in the future.

Licenser (approved):

Licenser:
I manually merged this to avoid the conflicts :) thank you!

PR #246, opened 2022-09-14
Fix/x86arch

universalmind303:
fixes #245

PR #376, opened 2024-06-26
Fix x86 compilation

PigeonF:
Compiling for the `i686-unknown-linux-gnu` target leads to multiple compilation failures.

The first set was easy to solve (akin to #246). The second issue is that `std::arch::x86_64::_mm_cvtsi128_si64` does not seem to have an equivalent for `std::arch::x86`. I think this is an oversight in `std::arch::x86`?

I am not confident enough in my SIMD knowledge to create a clever workaround, so I copied the "native rust implementation" for `x86`. If there is some workaround without using `_mm_cvtsi128_si64`, then feel free to change it.

> [!NOTE]
> I do not have a `x86` system myself, so I could not test the changes locally. I only checked by running `cargo build --target i686-unknown-linux-gnu`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,3 +146,11 @@ codegen-units = 1
 lto = true
 opt-level = 3
 panic = "abort"
+
+[lints.rust]
+unexpected_cfgs = { level = "allow", check-cfg = [
+    # Legacy code
+    'cfg(portable)',
+    # Tool specific configurations
+    'cfg(tarpaulin_include)',
+] }
--- a/src/value/tape/cmp.rs
+++ b/src/value/tape/cmp.rs
@@ -9,7 +9,7 @@ impl<'tape, 'input> PartialEq for Value<'tape, 'input> {
     #[cfg_attr(not(feature = "no-inline"), inline)]
     #[must_use]
     fn eq(&self, other: &Self) -> bool {
-        self == other
+        self.0 == other.0
     }
 }
 
--- a/src/impls/avx2/deser.rs
+++ b/src/impls/avx2/deser.rs
@@ -1,10 +1,10 @@
 #[cfg(target_arch = "x86")]
-use std::arch::x86::{
-    __m256i, _mm256_cmpeq_epi8, _mm256_loadu_si256, _mm256_movemask_epi8, _mm256_set1_epi8,
-    _mm256_storeu_si256,
-};
+use std::arch::x86 as arch;
+
 #[cfg(target_arch = "x86_64")]
-use std::arch::x86_64::{
+use std::arch::x86_64 as arch;
+
+use arch::{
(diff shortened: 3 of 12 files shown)


Licenser (approved):

Licenser:
nice catch and good solution. Using the native implementation is fine, 32-bit x86 systems are rare these days so I don't think it'll have a wide impact and with LLVM's optimisations it'll be plenty fast.

I suspect that the reason `_mm_cvtsi128_si64` doesn't exist in `x86` is that it returns a 64 bit value and the `x86` cpu's don't have registers to fit that into nicely.

The `arch` alias is a nice addition too, I never thought of that but it's much cleaner, I learned some new trick here :) thanks!

There are a few issues with the tests since we got a new clippy lints, those are not related to your changes I think. I'm happy to fix those in the main branch and rebase your PR tomorrow if you don't want to deal with them.

CI aside, PR looks great 👍 thanks!

PigeonF:
There is one more clippy lint that is reported on nightly: [`missing_transmute_annotations`](https://rust-lang.github.io/rust-clippy/master/index.html#/missing_transmute_annotations), specifically for

https://github.com/simd-lite/simd-json/blob/83a806f82155ed177807fa0af3e597737cd71a2d/src/macros.rs#L1253-L1259

I am not sure if you want to just ignore the lint here or not, so I have not done anything about them. Properly fixing the lint would probably involve removing the macros and calling `transmute` directly instead, which seems like a whole lot of effort. On the other hand, I am not familiar with the code base, so I am not comfortable deciding that the lint should be ignored 😅

Cargo.toml
```diff
@@ -146,3 +146,11 @@ codegen-units = 1
 lto = true
 opt-level = 3
 panic = "abort"
+
+[lints.rust]
+unexpected_cfgs = { level = "allow", check-cfg = [
+    # Legacy code
+    'cfg(portable)',
```
PigeonF, line 153: I am not too sure about the `portable` config (i.e. is the config supposed to be "hidden", or is this just remainder from previous code), so I have elected to ignore the warning about it here,.

src/value/tape/cmp.rs
```diff
@@ -9,7 +9,7 @@ impl<'tape, 'input> PartialEq for Value<'tape, 'input> {
     #[cfg_attr(not(feature = "no-inline"), inline)]
     #[must_use]
     fn eq(&self, other: &Self) -> bool {
-        self == other
+        self.0 == other.0
```
PigeonF, line 12: I believe the previous implementation was buggy, the clippy lint indicates that it would have looped infinitely.

I think comparing the wrapped value is what was intended here?

Licenser: Yes that's right! Good catch :)

I stumbled about that the other day too, clippy is my secret hero to be honest 😂 it avoids so many issues.

Licenser:
Perfect :) thank you  so much!

Licenser:
Sorry it took me so long to merge this, I thought I had done it and must have forgotten.

PR #384, opened 2024-08-23
Fix int parsing

### LINKED ISSUE #383: Untagged enums incorrectly deserialized in some edge cases
Hi we had a issue where simd-json would fail to deserialize a specificc structure containing a handwritten deserialize implementation. But it only happens when it is inside of a untagged enum.

My best bet on the reason is that even though we call `deserialize_u16` it seems to attempt with a i64 which is a error because on the default implementation and therefore it fails.

I have made a reproducible example:
<details>

`src/main.rs`:

```rust
use serde::{
    de::{Error as DeError, Visitor},
    Deserialize, Deserializer,
};
use std::fmt::{Formatter, Result as FmtResult};

const JSON: &str = r#"{
    "channels": [
      {
        "default_auto_archive_duration": 10080
      }
    ],
    "unavailable": false
}"#;

fn main() {
    unsafe {
        let mut json = JSON.to_string();
        let a = dbg!(simd_json::from_str::<Guild>(&mut json).unwrap());
        let b = dbg!(simd_json::from_str::<UnavailableGuild>(&mut json).unwrap());
        let c = dbg!(simd_json::from_str::<GuildCreate>(&mut json).unwrap());
        assert_eq!(a, serde_json::from_str::<Guild>(&mut json).unwrap());
        assert_eq!(b, serde_json::from_str::<UnavailableGuild>(&mut json).unwrap());
        assert_eq!(c, serde_json::from_str::<GuildCreate>(&mut json).unwrap());
    };
}

#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
#[serde(untagged)]
pub enum GuildCreate {
    Available(Guild),
    Unavailable(UnavailableGuild),
}

#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq)]
pub struct UnavailableGuild {
    pub unavailable: bool,
}

#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Hash)]
pub struct Guild {
    pub channels: Vec<Channel>,
}

#[derive(Clone, Debug, Deserialize, Eq, Hash, PartialEq)]
pub struct Channel {
    pub default_auto_archive_duration: Option<AutoArchiveDuration>,
}

#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
// Fix 1: Uncomment this and comment the handwritten deser impl.
// #[derive(Deserialize)]
// #[serde(from = "u16")]
pub enum AutoArchiveDuration {
    Hour,
    Day,
    ThreeDays,
    Week,
    Unknown { value: u16 },
}

impl AutoArchiveDuration {
    pub const fn number(self) -> u16 {
        match self {
            Self::Hour => 60,
            Self::Day => 1440,
            Self::ThreeDays => 4320,
            Self::Week => 10080,
            Self::Unknown { value } => value,
        }
    }
}

impl From<u16> for AutoArchiveDuration {
    fn from(value: u16) -> Self {
        match value {
            60 => Self::Hour,
            1440 => Self::Day,
            4320 => Self::ThreeDays,
            10080 => Self::Week,
            value => Self::Unknown { value },
        }
    }
}

impl<'de> Deserialize<'de> for AutoArchiveDuration {
    fn deserialize<D: Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
        deserializer.deserialize_u16(U16EnumVisitor).map(u16::into)
    }
}

pub struct U16EnumVisitor;

impl<'de> Visitor<'de> for U16EnumVisitor {
    type Value = u16;

    fn expecting(&self, f: &mut Formatter<'_>) -> FmtResult {
        f.write_str("u16")
    }

    fn visit_u16<E: DeError>(self, value: u16) -> Result<Self::Value, E> {
        Ok(value)
    }

    fn visit_u64<E>(self, v: u64) -> Result<Self::Value, E>
    where
        E: DeError,
    {
        v.try_into().map_err(DeError::custom)
    }

    // Fix 2: Uncomment this
    // fn visit_i64<E>(self, v: i64) -> Result<Self::Value, E>
    // where
    //     E: DeError,
    // {
    //     v.try_into().map_err(DeError::custom)
    // }
}
```

`Cargo.toml`:

```toml
[package]
name = "simdjsonbug"
version = "0.1.0"
edition = "2021"

[dependencies]
serde = { version = "1.0.208", features = ["derive"] }
serde_json = { version = "1.0.125" }
simd-json = { version = "0.14.0-rc.2" } # Also happens for 0.13.
```

</details>

Licenser:
We falsly returned i64 for non negative numbers that were in the i64 range; fixes #383

### DIFF
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -100,6 +100,7 @@ where
 {
     let mut deserializer = stry!(Deserializer::from_slice(s.as_bytes_mut()));
 
+    dbg!(&deserializer.tape);
     T::deserialize(&mut deserializer)
 }
 
@@ -222,7 +223,8 @@ impl<'de> Deserializer<'de> {
     #[cfg_attr(not(feature = "no-inline"), inline)]
     #[allow(clippy::cast_sign_loss)]
     fn parse_u16(&mut self) -> Result<u16> {
-        match unsafe { self.next_() } {
+        let next = unsafe { self.next_() };
+        match next {
             Node::Static(s) => s
                 .as_u16()
                 .ok_or_else(|| Self::error(ErrorType::ExpectedUnsigned)),
--- a/src/serde.rs
+++ b/src/serde.rs
@@ -100,7 +100,6 @@ where
 {
     let mut deserializer = stry!(Deserializer::from_slice(s.as_bytes_mut()));
 
-    dbg!(&deserializer.tape);
     T::deserialize(&mut deserializer)
 }
 
--- a/src/numberparse/correct.rs
+++ b/src/numberparse/correct.rs
@@ -191,11 +191,12 @@ impl<'de> Deserializer<'de> {
         } else if is_structural_or_whitespace(get!(buf, idx)) == 0 {
             err!(idx, get!(buf, idx))
         } else {
-            Ok(StaticNode::I64(if negative {
-                unsafe { static_cast_i64!(num.wrapping_neg()) } // -(num as i64)
+            Ok(if negative {
+                StaticNode::I64(unsafe { static_cast_i64!(num.wrapping_neg()) })
+            // -(num as i64)
             } else {
-                num as i64
-            }))
+                StaticNode::U64(num)
+            })
         }
     }
 }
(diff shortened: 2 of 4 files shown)


src/serde.rs
```diff
@@ -100,6 +100,7 @@ where
 {
     let mut deserializer = stry!(Deserializer::from_slice(s.as_bytes_mut()));
 
+    dbg!(&deserializer.tape);
```
Gelbpunkt, line 103: I assume this is a testing leftover?

Licenser: ah yes! I totally not wanted to put that into every deserialisation :D thank you

Issue #391, opened 2024-09-17
Latest stable (0.13.10) has a dependency on an unstable version of lexical-core


bassmanitram:
`lexical-core` 0.8 has the following advisory against it https://rustsec.org/advisories/RUSTSEC-2023-0086.

Licenser:
cherry picked your update into the v0.13 branch and released

PR #392, opened 2024-09-17
Update lexical-core version

bassmanitram:
Due to https://rustsec.org/advisories/RUSTSEC-2023-0086

This needs applying 0.13.10 - not sure how to properly manage the creation of the PR for that.

#391


Licenser:
heya, github seems to really not like me today I've gone and release 0.14 which completely removes lexical :)

Issue #398, opened 2024-10-19
owned::Value and borrowed::Value don't implement std::cmp::Eq


bassmanitram:
My continuing attempts to get [simd_json "dropped in" as a replacement for serde_json](https://github.com/bassmanitram/aws-lambda-rust-runtime/tree/feature/support-simd-json) (with some inevitable mutability hacks of course) are stymied at the moment because, while serde::Value implements std::cmp::Eq, the simd_json equivalents do not. Is there a reason for this, or is it simply and oversight?

Licenser:
This concerns floats (the bane of everyone's existence). Floats notoriously do not offer EQ since numbers that should be equal don't always turn out equal but might end up in a tolerance. On the other hand, numbers in that tolerance might not be equal.

bassmanitram:
Well yes that would explain it... But, then again, what does serde_json do
to 'get away with it' that simd_json couldn't do (somewhat rhetorical
since, if you don't know, I'll look into it)

bassmanitram:
And there it is - in serde_json Numbers are stored as strings until they are used via "is_xx" and "and_xx" - so Value CAN be Eq. This actually seems to me to be very sensible, since, when encapsulated by Value we are looking at "JSON" values which are always strings. They only become "float" when the caller says "can you be a float, if so gimme a float".

Looking at the simd_approach, that doesn't work - you assume that the user wants the float when you create the Value rather than when the Value is actually "unwrapped". So your comment above is correct - floats can't be strictly Eq. Unfortunately that's a significant blocker to having simd_json be a "fairly easy" droppin replacement for serde_json.

I'm guessing that modifying the approach to Number storage is a no-go ... what about a feature "serde_json" that changes how OwnedValue is implemented ... oh... or a third "type" of Value ... one that is compatible with serde_json (I could work on either of those, just don't want to "waste my time" if you guys don't think it's feasible or desirable).

bassmanitram:
OR ... could we store the string value with the float value in order to handle the "equivalence" constraint?

Kriskras99:
If you just want Eq, you could also just use the f32::total_cmp function (stable since Rust 1.62.0).
The issue I see, is that multiple string representations can give the same float which means Eq would say two fields are equal while they aren't.
Maybe add a feature flag for Eq, so people can opt-in/out

Licenser:
Ja strings are not a great option but @bassmanitram's PR for value-trait looks like a really good way to get the best of everything without the need to compromise.

[...]

PR #399, opened 2024-10-22
Feature/ordered float

bassmanitram:
This PR addresses issue #398 by implementing a feature named `ordered-float` that activates the same-named feature in [`value-trait`](https://github.com/simd-lite/value-trait/pull/54) which itself activates the wrapping of floats with [`OrderedFloat`](https://docs.rs/ordered-float/4.4.0/ordered_float/struct.OrderedFloat.html) and thus allows `StaticValue` to be `std::cmp::Eq`. Various feature-related "tweaks" are also implemented, while trying to use as few `#cfg[feature...]` annotations as possible.

Obvious the PR is dependent upon the `value-trait` PR, but while that is pending `Cargo.toml` in this PR's source branch patches the `value-trait` crate source to be the source repo and branch of the `value-trait` PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ name = "to_tape"
 harness = false
 
 [features]
-default = ["swar-number-parsing", "serde_impl", "runtime-detection", "ordered-float"]
+default = ["swar-number-parsing", "serde_impl", "runtime-detection"]
 
 arraybackend = ["halfbrown/arraybackend"]
 
@@ -155,7 +155,3 @@ unexpected_cfgs = { level = "allow", check-cfg = [
     # Tool specific configurations
     'cfg(tarpaulin_include)',
 ] }
-
-
-[patch.crates-io]
-value-trait = { path = '/home/jbartle9/Development/git/repositories/personal/value-trait' }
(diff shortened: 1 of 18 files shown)


bassmanitram:
This is a draft PR at the moment. However, at the time of this comment all tests pass with and without the new feature, and all _my_ clippy lints have been addressed.

Licenser:
value-trait is updated to .10.1 with your ordered-float

Licenser:
can we add the feature to the test matrix so we have the ordered float path covered?

bassmanitram:
> can we add the feature to the test matrix so we have the ordered float path covered?

Your wish is my command (which I was actually doing when you posted that :) )

Licenser:
great minds think alike :)

Licenser (approved):
Awesome improvement!

bassmanitram:
The goal of solving the Eq issue in my main project has been achieved - still a couple of issues that will likely be best dealt with in that project rather than hear (the necessary unsafe and mutability differences and a couple of missing methods only used in tests)

Once I get those tweaked away, I'll see if my goal is achieved.

Licenser:
Awesome, I think this needs a cargo fmt to  make the CI happy; otherwise good to merge :) if I can help with some thoughts on other problems feel free to ping

PR #403, opened 2024-10-24
Serde json error classification

bassmanitram:
(Again a "make simd_json drop more easily into a serde_json context") these methods on Error _roughly_ classify the error in the same way as their namesakes on `serde_json::Error`.

Cargo clippy, fmt, and test all run.


Licenser (approved):
I love it :)

PR #404, opened 2024-10-26
Implement a similar feature to serde_json's "RawValue"

bassmanitram:
This is not exactly like RawValue but allows some of the same tricks - i.e. get back a "raw" representation of an initially-parsed and verified JSON stream. simd_json does it with the Tape while serde_json parses the string as it goes.

So THIS aims to return the Tape from the Deserializer without having to cast the Deserializer trait into a simd_json::Deserializer (which is nigh-on impossible anyway). The technique is similar to that implemented by RawValue, but is simpler because our goals are a little less all-encompassing than those of RawValue.

The goal for _me_ is to be able to get into a Deserialize implementation, extract the tape, peek into it to select a candidate type to deserialize, then deserialize the Tape to that type.

The serde_json RawValue use-case that I'm trying to simd_json-ize can be found [here](https://github.com/awslabs/aws-lambda-rust-runtime/blob/19c190d6762ea755cb14462b6c0347afb4352a29/lambda-http/src/deserializer.rs#L18).

### DIFF
--- a/src/serde/de.rs
+++ b/src/serde/de.rs
@@ -4,6 +4,8 @@ use serde_ext::de::{self, DeserializeSeed, MapAccess, SeqAccess, Visitor};
 use serde_ext::forward_to_deserialize_any;
 use std::str;
 
+pub const TAPE_TOKEN: &str = "$simd_json::private::Tape";
+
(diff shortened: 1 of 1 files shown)


bassmanitram:
Closing this for now since there is a small mod that should go in first before this work can really get underway.

PR #405, opened 2024-10-28
Deserializer as Value and restartable

bassmanitram:
This is kind of a prereq for any RawValue-like capability. 

As ever, 'my' use case requires the Deserialize to "introspect" or "peek at" the JSON, possibly multiple times, in order to "try" possibly multiple different "sub deserializations" for an Enum. With `serde_json::RawValue`, you have access to the string slice that backs the value, so the code tries that string as various different types (enabled by features). 

Because `simd_json` doesn't seem to keep the string around (and, anyway, it's a mutated bytes buf), the alternative is to introspect the Node vector held by the deserializer. Hence `as_value`. 

Additionally, reusing the Deserializer - which multiple attempted deserializations would require - won't work without some sort of "start again" operation since a deserialization moves the "cursor" (updates the internal index) ... hence `restart`.

in _my_ use case this (almost unbelievably) works! 

Now, safely getting _access_ to the type-specific methods of the deserializer passed into `Deserialize::deserialize` ... well, that's where the _future_ RawValue feature will come in. At the moment, though, I do it with some wildly unsafe casting (which is actually ALMOST legitimate in my case because there is "no way" the Deserializer _couldn't_ be a simd_json::Deserializer at the point  I'm do the cast - famous last words, eh!).

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -213,7 +213,7 @@ impl Error {
         match &self.error {
             ErrorType::Io(_) | ErrorType::InputTooLarge => true,
             ErrorType::InternalError(e) if !matches!(e, crate::InternalError::TapeError) => true,
-            _ => false
+            _ => false,
         }
     }
 
@@ -234,7 +234,8 @@ impl Error {
     #[must_use]
     pub fn is_syntax(&self) -> bool {
         // Lazy? maybe but if it aint something else...
-        matches!(self.error,
+        matches!(
+            self.error,
             ErrorType::InternalError(crate::InternalError::TapeError) | //This seems to get thrown on some syntax errors
             ErrorType::BadKeyType |
             ErrorType::ExpectedArrayComma |
@@ -258,7 +259,8 @@ impl Error {
             ErrorType::ExpectedObjectContent |
             ErrorType::ExpectedObjectKey |
             ErrorType::Overflow |
-            ErrorType::SimdUnsupported)
+            ErrorType::SimdUnsupported
+        )
     }
 }
 impl std::error::Error for Error {}
--- a/src/error.rs
+++ b/src/error.rs
@@ -211,7 +211,7 @@ impl Error {
     pub fn is_io(&self) -> bool {
         // We have to include InternalError _somewhere_
         match &self.error {
-            ErrorType::Io(_) => true,
+            ErrorType::Io(_) | ErrorType::InputTooLarge => true,
             ErrorType::InternalError(e) if !matches!(e, crate::InternalError::TapeError) => true,
             _ => false
         }
@@ -236,7 +236,6 @@ impl Error {
         // Lazy? maybe but if it aint something else...
         matches!(self.error,
             ErrorType::InternalError(crate::InternalError::TapeError) | //This seems to get thrown on some syntax errors
-            ErrorType::InputTooLarge |
             ErrorType::BadKeyType |
             ErrorType::ExpectedArrayComma |
             ErrorType::ExpectedObjectColon |
--- a/src/error.rs
+++ b/src/error.rs
@@ -219,17 +219,42 @@ impl Error {
         matches!(self.error, ErrorType::Eof)
     }
 
-    /// Indicates if the error that occurred was due to a deserializer error
+    /// Indicates if the error that occurred was due to a data shape error
     #[must_use]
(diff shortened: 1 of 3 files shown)


bassmanitram:
At the moment there are no tests - working on those.

@Licenser - this may also not be entirely safe - we could certainly mark either new method as unsafe if you prefer. Also I was 'forced' to open up the internals of the Deserializer to the Crate rather than just the super module - please check that this is OK.

Licenser:
Okay, I'm reading things out of order in my github notifications :D;

I'd like to suggest an alternative path: implement Deserialize for Value (perhaps use that in the deserializer, too, to avoid double implementations). The tape value has zero overhead since it's just a wrapper, so we can always create it without having to worry. The example from the other PR would become something like this(like because I never can remember the deserializer API o.O):

```
impl<'de> Deserialize<'de> for LambdaRequest {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        let tape = deserializer.into_value()?;

        #[cfg(feature = "apigw_rest")]
        if let Ok(res) = ApiGatewayProxyRequest::deserialize(tape) {
            return Ok(LambdaRequest::ApiGatewayV1(res));
        }
        #[cfg(feature = "apigw_http")]
        if let Ok(res) = ApiGatewayV2httpRequest::deserialize(tape) {
            return Ok(LambdaRequest::ApiGatewayV2(res));
        }
        #[cfg(feature = "alb")]
        if let Ok(res) = AlbTargetGroupRequest::deserialize(tape) {
            return Ok(LambdaRequest::Alb(res));
        }
        #[cfg(feature = "apigw_websockets")]
        if let Ok(res) = ApiGatewayWebsocketProxyRequest::deserialize(tape) {
            return Ok(LambdaRequest::WebSocket(res));
        }
        #[cfg(feature = "pass_through")]
        if PASS_THROUGH_ENABLED {
            return Ok(LambdaRequest::PassThrough(data.json_string()));
        }

        Err(Error::custom(ERROR_CONTEXT))
    }
}
```

bassmanitram:
Absolutely I intend to implement Deserialize for Value - that's the
"similar feature to RawValue" that I am referring to. I'll review the need
for "restart" but in testing here I found I needed it because
deserialization advances the tape so when a later deserialization starts,
part of the data has already been "consumed". That "restart" method fixes
it.

bassmanitram:
Just made a test without restart... seems it ISN'T required for my
value-based introspection ... but I'm pretty sure it would be if I was
trying to reuse the Deserializer. I'll include that check in my tests
(which, now I have all the other stuff working, I'm onto)

bassmanitram:
Tests added and passing - see the "restart" test for the use case.

Licenser:
Reading this, and thinking about things gives me one big yuck. having `reset` will make nested deserializer impossible and possibly silently break them as a deserializer is not guaranteed to start from scratch. Let me give an example


```rust

enum Resetter {...}

impl Deserialize for Resetter {
... do something with reset ...
}


let r: Resetter = simde_json::from_str(...) // yay it works

#[derive(Deserialize)]
struct Wrapped {
  r: Resetter,
  s: String 
}

let w: Wrapped = simde_json::from_str(...) // boom it breaks, resetter will reset the deserializer for wrapper
```

bassmanitram:
> Reading this, and thinking about things gives me one big yuck. having `reset` will make nested deserializer impossible and possibly silently break them as a deserializer is not guaranteed to start from scratch. Let me give an example
> 
> ```rust
> enum Resetter {...}
> 
> impl Deserialize for Resetter {
> ... do something with reset ...
> }
> 
> 
> let r: Resetter = simde_json::from_str(...) // yay it works
> 
> #[derive(Deserialize)]
> struct Wrapped {
>   r: Resetter,
>   s: String 
> }
> 
> let w: Wrapped = simde_json::from_str(...) // boom it breaks, resetter will reset the deserializer for wrapper
> ```

Understood ... so perhaps we mark as unsafe? Actually, in my use case, I'm not gonna need restart now ... so I could just remove it?

Licenser:
I think removing it is good, given it would potentially make things outside of the scope of the implementer explode it'd be hard to trickle the unsafe up to the user so they'd never know

PR #408, opened 2024-11-23
Remove line about being tested with Miri

Arzte:
Miri has been disabled since August 2023 (https://github.com/simd-lite/simd-json/commit/a03d8f979e095b36ee1a16ff280e21a6ef4b9933) with no obvious movement to re-enable testing. Removing it as a tagline feature will help to remove any uncertainty or bad feelings about this project unintentionally lying about miri UB testing.


Licenser:
Good catch, thanks! I should have removed that when the mini tests had to go.

PR #415, opened 2025-02-24
Remove a starting underscore in floating point exponent

petrochenkov:
This is a questionable feature in Rust lexer and `simd-json` is the only known crate on crates.io that is using it. It's not being removed *yet*, but it would still be better to avoid using it.

See https://github.com/rust-lang/rust/pull/137394 and linked issues for more details.


Licenser:
Thank you! Good call :)

PR #425, opened 2025-04-23
Update README.md

darach:
Update the allocator text as snmalloc is in common use for simd-json with tremor and has been for years. Also, most of the goodness in mimalloc has since been merged into snmalloc. Rephrase accordingly.


Licenser (approved):
Really good catch, thanks!

darach:
I'm surprised we didn't catch this sooner. I mean, we've using snmalloc now for years! 🤷

Licenser:
yea I guess we don't often read our own readies 😅

PR #428, opened 2025-07-03
examples/perf: Add an arch flag to run only on x86_64

Sylfrena:
Allow lints for "unnecessary transmutes" for now.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -1234,7 +1234,7 @@ pub(crate) use static_cast_i32;
 macro_rules! static_cast_u32 {
     ($v:expr_2021) => {
         // #[allow(clippy::missing_transmute_annotations)]
-        ::std::mem::transmute::<i32, u32>($v)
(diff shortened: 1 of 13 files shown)


src/macros.rs
```diff
@@ -1223,7 +1223,7 @@ pub(crate) use unlikely;
 #[allow(unused_macros)]
 macro_rules! static_cast_i32 {
     ($v:expr_2021) => {
-        ::std::mem::transmute::<u32, i32>($v)
+        u32::cast_signed($v.wrapping_neg())
```
Licenser, line 1226: The wrapping_neg should definitely not be here. The macro name gives no indication that it also negates the number. This would be highly misleading and potentially lead to bugs down the road.

Also `cast_signed` requires a newer rust version which might be OK but needs careful consideration as it is a backwards compatibility breaking change.

Licenser (approved):
Thank you so much!

PR #438, opened 2025-10-12
docs: add a note about endianness

### LINKED ISSUE #437: docs: note that simd-json doesn't work on big endian platforms
In my project - qsv, I had to put in conditional compilation directives to use serde_json in big endian platforms (e.g. s390x) and only use simd-json, on little endian.

Perhaps, a short note in the README.md?

jqnatividad:
resolves #437


Licenser:
little endian platforms should now work correclty

PR #442, opened 2026-02-24
fix: alloc(0) call is UB

EmilyMatt:
When calling alloc(), the layout must be non-zero, otherwise we fall into undefined behaviour.
This is documented in the function's docs:
https://doc.rust-lang.org/beta/src/core/alloc/global.rs.html#152

We can just make a dangling pointer for those cases, and not deallocate it

Found by running miri on tests::test_send_sync

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -125,7 +125,7 @@ jobs:
           targets: wasm32-wasip1
 
       - name: Install wasmtime
-        run: cargo install wasmtime-cli
+        run: cargo install wasmtime-cli --version 38.0.4
 
       - name: Run tests
         run: |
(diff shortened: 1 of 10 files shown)


Licenser (approved):
Thank you! good catch :)

EmilyMatt:
@Licenser Does this have anything actionable for me? or is just pending due to time limitations?

EmilyMatt:
I'm fixing CI here(some clippy warnings and wastime installation)
Will rebase the other PR (#445 ) after this is merged

EmilyMatt:
@Licenser Last two checks can't be green until merged to main (I think they will only be green in the next PRs)

Licenser:
Heya @EmilyMatt, sorry for the long delay, live has been super busy and I have had every little time for open source. I'm trying to catch up a little 😅 and get your PR's merged.

PR #443, opened 2026-02-24
fix: Miri error on invalidated mut borrow

EmilyMatt:
Because the function continuously mutates the data in the input mut slice, miri's borrow checker invalidates the original mut borrow, so calling input.as_mut_ptr() is not allowed(miri no longer guarantees the validity of `input`)
This can be fixed by calling input.as_mut_ptr()  once and using that var


Licenser (approved):
good catch, very nice improvement, thank you!

PR #444, opened 2026-02-24
fix: parse_eight_digits_unrolled requires SSSE3 but isn't marked

EmilyMatt:
swar-number-parsing requires ssse3(due to using ssse3 calls), but does not mark it as a requirement, ergo we can reach invalid instruction calls on x86 systems that don't have ssse3.

Needed to add a fallback as well as rename the function and mark it ssse3

Found using miri on  numberparse::correct::test::float


Licenser:
Thank you!

PR #445, opened 2026-02-24
fix: No UTF-8 validation happening when `portable` is disabled

EmilyMatt:
When runtime-detection is enabled, but portable is not, we enter a code path that does *NOT* validate the UTF-8 input.
We then continue normally, reaching a codepath that claims that the UTF-8 was validated in the previous call(which it is, except not for the not(feature = "portable") code path

Added a fairly naive validation here, open to other suggestions.

Found by running miri on jsonchecker fail34 

PR #442 from the same bunch will fix the CI issues, so this should be rebased and merged after

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -125,7 +125,7 @@ jobs:
           targets: wasm32-wasip1
 
       - name: Install wasmtime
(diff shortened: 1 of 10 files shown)


Licenser (approved):
Yay!

Issue #89, opened 2019-12-18
Improve code coverage for integer errors


Licenser:
We have good coverage for the happy case on numbparse but we do not check for errors nicely - we could improve here.

PR #465, opened 2026-07-12
Add integer parsing error-path test coverage

mag123c:
Adds test coverage for integer parse errors, see #89.

`parse_number` had solid happy-path coverage but most of its `InvalidNumber` branches were untested. Added tests for:

- leading zeroes — `01`, `-01`
- base prefixes — `0x1`, `0b1`, `0o1`
- a `-` not followed by a digit — `-`, `-x`
- an exponent marker with no exponent digits — `1e`, `1e+`
- non-terminator bytes after the digits — `123a`, `1_000`

Test-only, no parser changes. Covers the default (`correct.rs`) path; happy to mirror the same into the `approx-number-parsing` path too if you'd prefer.

### DIFF
--- a/src/numberparse/correct.rs
+++ b/src/numberparse/correct.rs
@@ -495,10 +495,63 @@ mod test {
 
     #[test]
     fn int_trailing_invalid() {
-        // An embedded NUL after the digits is not a valid terminator and must be
-        // rejected.
+        // The byte terminating an integer must be structural-or-whitespace. An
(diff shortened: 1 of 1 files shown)


Licenser (approved):
Nice! I like it more tests are more better!

Licenser:
If you want to add the same to the apron parsing 👍 happy to merge another PR

PR #466, opened 2026-07-13
Add approx integer parsing error-path test coverage

mag123c:
Follows up on #465 — the maintainer offered to merge the same coverage for the approx parser.

## What

- Add `InvalidNumber` error-path tests for the `approx-number-parsing` integer parser, mirroring the correct-parser coverage from #465.
  - Covers leading zeroes, base prefixes (`0x`/`0b`/`0o`), a bare/invalid sign, an exponent marker with no digits, and non-terminator trailing bytes.
- Add a `to_value_from_str` test helper matching the one in `correct.rs`.

## Notes

- Test-only change: `approx.rs`'s `parse_number` already rejected these inputs, the paths were simply untested.
- Verified with `cargo test --features approx-number-parsing` and `cargo fmt --check`.

### DIFF
--- a/src/numberparse/approx.rs
+++ b/src/numberparse/approx.rs
@@ -562,9 +562,17 @@ impl Deserializer<'_> {
 #[cfg(test)]
 mod test {
     #![allow(clippy::default_trait_access)]
+    use crate::error::Error;
+    use crate::value::owned::Value;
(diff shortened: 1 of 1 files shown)


Licenser (approved):
Yay!!!

PR #475, opened 2026-08-19
Added AVX512BW support

Marabii:
Hello maintainers,

I added AVX512BW support, a new implementation next to the others (avx2, ...)
It is automatically selected in runtime as the fastest solution when available just like before.

`flatten_bits` uses AVX-512VBMI2's `vpcompressb` when available, with a
 scalar fallback when it isn't.

I had to bump up `SIMDJSON_PADDING` from 32 to 64 to match the `__m512i` width.

MSRV bumped to 1.89 (required by the AVX-512 intrinsics), CI toolchains and README updated accordingly.

Benchmarked against AVX2 on a 550MB generated JSON file: ~5% faster on `to_tape` and roughly on par for full `to_value`, since that path is dominated by hashmap/allocation costs rather than stage-1 parsing

Full test suite ran perfectly.

Please let me know if there are any modifications needed before it can be merged.

Thanks!

### DIFF
--- a/.github/workflows/big-endian.yml
+++ b/.github/workflows/big-endian.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - uses: dtolnay/rust-toolchain@1.88
+      - uses: dtolnay/rust-toolchain@1.89
 
       - name: Install cross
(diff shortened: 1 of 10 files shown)


Marabii:
Why did no one respond to my PR ?

